PR #4, opened 2021-10-31
Input/atomics

### LINKED ISSUE #3: Better input
In a similar fashion to futurecoder, input should be gathered by overriding sys.stdin.readline instead of input itself and making it async, requiring a translation with a brittle regex.

winniederidder:
Closes #3 
Implementation of synchronous I/O calls (both for Python input/print and JavaScript prompt.

Requires SharedArrayBuffers and Atomics to allow the WebWorker to wait for the main thread, thus Safari is currently not supported. Teachers should thus ask their students to use Google Chrome/FireFox or something else listed on https://caniuse.com/?search=atomics. Regarding Safari, it is available in the Technology Preview version so hopefully support will be re-enabled soon.

As an extra, this PR also includes conversion of the WebWorkers to TypeScript and integration of Comlink to reduce the mental overhead of message passing. This however required ejecting (which would have to be done later anyway to get rid of all these React dependencies that exist solely to ease development).


bmesuere:
Can you explain why you need atomics and shared array buffers? More specifically, is there a reason why you're not simply using promises?

(in addition to Safari, Dodona also supports Firefox as early as version 62)

winniederidder:
> Can you explain why you need atomics and shared array buffers? More specifically, is there a reason why you're not simply using promises?
The relevant calls that require these SharedArrayBuffers are things like input and prompt, which are synchronous. Furthermore, these calls are being executed in a WebWorker so they cannot interact with users directly. In order to be able to let users input data on the main thread and then passing that to the worker thread to resume its execution requires blocking the thread at the relevant call. Blocking can be done reliably and efficiently using Atomics. 
Promises cannot be used because these calls are synchronous. An alternative I suggested is indeed turning these calls into promises themselves, but this approach is quite brittle. Calls to `input` would have to be replaced with `await input`, and functions using input would have to be replaced with asynchronous versions as well (recursively). 

A potential alternative is [unthrow](https://github.com/joemarshall/unthrow), but it only works for Python 3.8, whereas Pyodide runs Python 3.9.  A relevant [issue](https://github.com/pyodide/pyodide/issues/1503) on the Pyodide repository goes more in depth but is also currently unresolved.

bmesuere:
To be sure I'm understanding this right, can you walk me through what happens when input is called in python? Since this seems quite complex, the explanation might also be added to the PR start and your thesis.

What I had in mind was manually creating an unresolved promise (in the worker) when input was called for and awaiting for that. When the user provides input, it is passed to the worker through an event which then manually resolves the promise with the provided value.

Totally unrelated:
- Many of your dependencies seem dev-dependencies so should be added as such to the package file
- Your code contains many console logs. Since this is complex code, they might be useful for future debugging purposes. Commenting them doesn't seem scalable, so it might be an option to add a wrapped console.log call that check if a debug/verbose/env flag is set to true.

winniederidder:
> To be sure I'm understanding this right, can you walk me through what happens when input is called in python? Since this seems quite complex, the explanation might also be added to the PR start and your thesis.

The builtin input method is overwritten to instead call a JavaScript function. This function sends a message to the main thread that it requires input. Then it waits until the first value in a shared memory buffer is not set to 0 (inputMetaData[0]); this indicates that the next line is ready in a different shared buffer (inputTextArray) with length equal to inputMetaData[1]. It then reads this value and returns it to Python.

> What I had in mind was manually creating an unresolved promise (in the worker) when input was called for and awaiting for that. When the user provides input, it is passed to the worker through an event which then manually resolves the promise with the provided value.

We cannot just await this Promise. Input in Python is synchronous, so when we use an asynchronous callback to handle input, we cannot guarantee that the input will be ready when it is used in Python (since it is not awaited, for this we would need to await the result of the callback in Python. The user expects his code to block until it has received input, so the SharedMemory is required to have this synchronously blocking behaviour.
> 
> Totally unrelated:
> 
> * Many of your dependencies seem dev-dependencies so should be added as such to the package file

I'm pretty certain there is a lot in there that can be thrown away, as I started from create-react-app and then ejected to have  a couple of extra functionalities available such as worker-plugin and devServer headers for the SharedBuffers. Not quite sure yet how to properly clean it up.

> * Your code contains many console logs. Since this is complex code, they might be useful for future debugging purposes. Commenting them doesn't seem scalable, so it might be an option to add a wrapped console.log call that check if a debug/verbose/env flag is set to true.

Adding a wrapper sounds like a good idea. Debugging multi-threaded code tends to cause a few headaches along the way.

bmesuere:
I've read all of the related issues in the Pyodide and now fully understand the issue (and why we can't await a promise).

Given that
- the code added here is quite involved
- has limitations on browser usage and requires additional server-side headers to be set
- Pyodide wants to solve this problem as well

I suggest holding off on implementing this ourselves for now. For a first release, we can make the input non-interactive, e.g., provide in input area on the page that will be fed into the python process when `input` is called.

@pdawyndt @chvp What are your thoughts on this?

pdawyndt:
Agree that we shouldn't spend time on solving issues that Pyodide has in scope as well. Do we have any sight on their roadmap / estimate when they hope to have resolved this issue?

Can we live with browser limitation for now, instead of relying on non-interactive iput only?

alexmojaki:
> Agree that we shouldn't spend time on solving issues that Pyodide has in scope as well. Do we have any sight on their roadmap / estimate when they hope to have resolved this issue?

I'd say the discussion of this stuff started in February in https://github.com/pyodide/pyodide/issues/1219 and a PR to start dealing with it appeared in April in https://github.com/pyodide/pyodide/pull/1547 and seems to have gone stale. There's quite a bit of demand for solutions but they all have drawbacks. It doesn't feel like this will be resolved by Pyodide soon.

I'm also using SharedArrayBuffer and Atomics in https://futurecoder.io/ and it's working nicely. I'm not happy about Safari and the cross origin headers, but on the other hand:

- unthrow looks weird and spooky
- Pyodide in a web worker doesn't work in Safari anyway: https://github.com/pyodide/pyodide/issues/441
- SharedArrayBuffer is also needed for interrupting, and I don't know of any alternative in a web worker.

winniederidder:
> * Pyodide in a web worker doesn't work in Safari anyway: [Maximum call stack error when running pyodide within a web worker pyodide/pyodide#441](https://github.com/pyodide/pyodide/issues/441)

This seems to occur based on the used packages? I believe @pdawyndt tried a couple simple scripts and did not encounter this, but it is good to know that this issue exist, so thank you for bringing this to our attention.

> * SharedArrayBuffer is also needed for interrupting, and I don't know of any alternative in a web worker.

An approach that is implemented here is simply terminating the Worker and restarting it, which isn't ideal but does work without the SharedArrayBuffer. The ServiceWorker-implementation using synchronous xhr requests works relatively well (albeit being a bit complicated and likely slower) as a fallback to SharedArrayBuffer, so in theory they can be eliminated. This method does feel a lot more hacky to me.

alexmojaki:
> An approach that is implemented here is simply terminating the Worker and restarting it

Ah, it doesn't take long to restart the worker and reinitialize Pyodide?

> The ServiceWorker-implementation using synchronous xhr requests works relatively well

Nice! I originally suggested that on the Pyodide issue but I didn't end up using it myself, glad it helped you.

winniederidder:
> Ah, it doesn't take long to restart the worker and reinitialize Pyodide?
> 
It takes a couple seconds at this time, but a user usually terminates execution when it is clear something is wrong and then needs time to fix the issue anyway, so not being able to run instantly doesn't appear to be an issue at this time.


> Nice! I originally suggested that on the Pyodide issue but I didn't end up using it myself, glad it helped you.
Indeed, I got the idea from your comment. This project is my master thesis and I currently only reference FutureCoder in the related works section. If you have a more direct way that you would like to be referenced by, then do let me know as your project is a huge source of inspiration.

PR #23, opened 2021-11-16
Set up github actions

### LINKED ISSUE #22: Enhance GitHub actions
The used GitHub actions should be more flexible: being able to be run manually, as well as over various branches instead of just development.

winniederidder:
This pull request sets up 2 github actions:

- A deploy action that runs on every push to main or on demand (by adding the deploy label) and will deploy the branch to github pages.
- A lint action that runs on every push that lints and type checks the code.

Closes #22

### DIFF
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -1,10 +1,12 @@
 name: Deploy on GitHub Pages
 on:
-  push:
-    branches:
-      - development
+  workflow_dispatch:
+  pull_request:
+    types:
+      - labeled
 jobs:
   build-and-deploy:
+    if: ${{ github.event.action == 'labeled' && github.event.label.name == 'deploy' }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -1,6 +1,9 @@
 name: Deploy on GitHub Pages
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - main
   pull_request:
     types:
       - labeled
@@ -23,3 +26,9 @@ jobs:
         with:
           branch: gh-pages # The branch the action should deploy to.
           folder: build # The folder the action should deploy.
+      - name: Remove deploy label
+        uses: buildsville/add-remove-label@v1
+        with:
+          token: ${{secrets.GITHUB_TOKEN}}
+          label: deploy
+          type: remove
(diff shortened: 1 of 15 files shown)


.github/workflows/deploy-pages.yaml
```diff
@@ -1,11 +1,16 @@
 name: Deploy on GitHub Pages
 on:
-  push:
-    branches:
-      - development
+  workflow_dispatch:
+  pull_request:
+    types:
+      - labeled
```
bmesuere, line 6: I would add push-branches-main here as well

.github/workflows/deploy-pages.yaml
```diff
@@ -1,11 +1,16 @@
...
 jobs:
   build-and-deploy:
+    if: ${{ github.event.action == 'labeled' && github.event.label.name == 'deploy' }}
```
bmesuere, line 9: If you trigger this with a deploy label, you should also remove the label at the end of the run.

bmesuere (changes requested):

bmesuere:
Awaiting full tests and linting, a second action could be added that runs the type checker and builds the project as poor man's test for each pull request.

.github/workflows/deploy-pages.yaml
winniederidder: The Dodona action uses a GitHub token for the remove label action. However, since I have no access to the repository settings, I cannot generate this access token myself. I will include it in this PR, can you add the secret?

bmesuere: No, GITHUB_TOKEN is set automatically for the current user.

winniederidder:
Whilst adding the Linting action, I of course had to fix all exisiting linting errors. I apologize for making reviewing a bit more difficult, but no functionality should have changed.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,27 @@
+name: Lint
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - '*'
+  pull_request:
```
bmesuere, line 9: This can be simplified to just run on push

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,27 @@
...
+      uses: actions/setup-node@v2
+      with:
+        node-version: 12.x
+        cache: yarn
+    - name: Install dependencies
+      run: |
+        yarn install
+    - name: Lint
+      run: |
+        yarn lint --quiet
```
bmesuere, line 27: Is there a specific reason to use `--quiet`?

bmesuere (changes requested):

.github/workflows/lint.yaml
winniederidder: I copied this from the Dodona lint action, but I can omit this.

bmesuere (approved):

PR #30, opened 2021-11-18
Indicate program state and add tailwindcss

### LINKED ISSUE #21: Add indicator for program state
Show a visual indication of what Papyros is doing, e.g.:
Running code (and maybe add a display when complete about runtime)
Awaiting input from the user
Terminating the current process

winniederidder:
This pull requests adds a spinner and textual state indicator of papyrus. Examples of states are loading, running and awaiting input.

In addition, the tailwind css library was added to help with the styling.

image
image

Closes #21


postcss.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+    plugins: [
+      require('postcss-import'),
+      require('tailwindcss/nesting'),
+      require('tailwindcss'),
+      require('autoprefixer'),
+      require("postcss-flexbugs-fixes"),
+      require("postcss-preset-env")({
+        autoprefixer: {
+          flexbox: "no-2009"
```
bmesuere, line 10: can't autoprefixer use the browserslist?

build/index.html
```diff
@@ -30,29 +30,43 @@
...
+      </div>
+      <div class="flex flex-row">
+        <button id="run-code-btn" type="button" class="mr-1 btn btn-primary">Run</button>
+        <button id="terminate-btn" type="button" class="mr-1 btn btn-danger" hidden>Terminate</button>
+        <div>
+          <svg id="state-spinner" class="animate-spin mr-3 h-5 w-5 text-white" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" display="none">
+            <circle class="opacity-25" cx="12" cy="12" r="10" stroke="red" stroke-width="4"></circle>
+            <path class="opacity-75" fill="red" d="M4 12a8 8 0 018-8V0C5.373 0 0 5.373 0 12h4zm2 5.291A7.962 7.962 0 014 12H0c0 3.042 1.135 5.824 3 7.938l3-2.647z"></path>
+          </svg>
+          <p id="application-state-p">Loading</p>
```
bmesuere, line 64: Is there a reason for using a `p` here? An inline element probably looks better:
image

src/Papyros.ts
```diff
@@ -125,18 +136,25 @@ export function papyros(inputTextArray?: Uint8Array, inputMetaData?: Int32Array)
...
         try {
+            stateSpinner.style.display = "";
+            stateP.innerText = "Running";
             await backend.runCode(codeArea.value);
         } catch (error: any) {
             onError(error);
         } finally {
+            const end = new Date().getTime();
+            stateSpinner.style.display = "none";
+            stateP.innerText = `Time taken to run: ${end - start} ms`;
```
bmesuere, line 149:
```suggestion
            stateP.innerText = `Code executed in ${end - start} ms`;
```

bmesuere (changes requested):
Does this include the full tailwind library in the final build, or is [unused css](https://tailwindcss.com/docs/optimizing-for-production#removing-unused-css) removed?

I have updated the PR start to include some screenshots and the mention of tailwind.

postcss.config.js
winniederidder: I took this from the Dodona postcss.config.js, so I assumed there is a good reason why it is present (as this is supposed to work with Dodona, so the same config should help).

bmesuere: I don't think we actually use postcss in Dodona (it might have been added by webpacker, but all css is actually handled by sprockets)

winniederidder: https://github.com/dodona-edu/dodona/blob/develop/postcss.config.js is indeed a rather old file. Which parts of this config file do you think are irrelevant then?

bmesuere: All of them. No settings are needed. By default the browserslist file is used

winniederidder: Removed it then, thank you!

bmesuere (approved):
:+1:

PR #31, opened 2021-11-18
Use uuid to ignore outdated events

### LINKED ISSUE #18: Code termination doesn't always work properly
Running the code

```python
name = input()
while name != 'stop':
    print(f'Hello, {name}!')
```

with input (`stdin`)

```
Alice
Bob
Claire
stop
```

Ends in an infinite loop. Clicking "Terminate" does not halt output generation and the "Run" button stays disabled. Can only be stopped by refreshing the page.

winniederidder:
Closes #18 
By using an incremental id attached to a code execution, we can filter out late messages sent by interrupted workers and allowing termination to work more reliably.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,9 @@
     "webpack-dev-server": "3.11.1"
   },
   "dependencies": {
-    "comlink": "^4.3.1"
+    "@types/uuid": "^8.3.3",
+    "comlink": "^4.3.1",
+    "uuid": "^8.3.2"
   },
   "scripts": {
     "start": "webpack serve",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -26,7 +26,9 @@
     "webpack-dev-server": "3.11.1"
   },
   "dependencies": {
-    "comlink": "^4.3.1"
+    "@types/uuid": "^8.3.3",
```
bmesuere, line 29: As far as I'm aware, types should always be dev dependencies.

bmesuere (changes requested):
Is there a reason for using UUIDs over simple incremental id's?

bmesuere (approved):

PR #32, opened 2021-11-19
Fix running code containing Python doctests

### LINKED ISSUE #16: Allow Python doctests to work
Python doctests don't work in Papyros. First of all the global variable `__name__` is `'builtins'`. But even then,

```python
if __name__ == 'builtins':
    import doctest
    doctest.testmod()
```

doesn't do its job.

winniederidder:
Closes #16
This PR allows doctests to work by doing the following:\
setting the module name to __main__ instead of builtins\
Changing the way the scoping works to allow doctest to find the relevant docstrings. This requires running the code in the global scope, and then resetting the scope to undo changes (preventing declared variables from being found later, as well as undoing overrides of builtins)\
Better conversion of PyProxy objects to actual JS objects to prevent type errors.

### DIFF
--- a/src/workers/python/PythonWorker.worker.ts
+++ b/src/workers/python/PythonWorker.worker.ts
@@ -50,8 +50,10 @@ class PythonWorker extends Backend {
         await this.pyodide.loadPackagesFromImports(code);
         if (this.initialized) {
             // run the code, potentially polluting the namespace
+            // Functions and variables defined by the user become global
+            // We need them to be global to let doctest work out of the box
             const result = this.pyodide.runPython(code);
-            // Find newly added globals
+            // Find these newly added globals
             const pyodideGlobals = this.pyodide.globals;
             const keysToRemove: Array<string> = [];
             for (const key of pyodideGlobals.keys()) {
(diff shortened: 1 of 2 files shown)


bmesuere (commented):
My Python knowledge isn't good enough to review these changes. Forwarding this to @pdawyndt and @chvp.

pdawyndt (approved):

src/workers/python/PythonWorker.worker.ts
```diff
@@ -3,51 +3,68 @@ import { Backend } from "../../Backend";
...
+                if (!this.globals.has(key)) {
+                    keysToRemove.push(key);
+                } else {
+                    // Reset value in case it was overriden
+                    pyodideGlobals.set(key, this.globals.get(key));
+                }
+            }
+            // remove them from the actual globals
+            keysToRemove.forEach(k => pyodideGlobals.delete(k));
+            return result;
```
chvp, line 67: Some comments here on what globals are added while running the code and why this is a problem would be nice. Is it only the `__name__` in `__override_builtins`, or are there other globals we want to remove again. Does this not also remove any functions defined by the user, and if so, is this something we want? (If it is, comment that they are and why.)

chvp (changes requested):

src/workers/python/PythonWorker.worker.ts
winniederidder: The code of the user is executed as a module, and functions/variables at the top level thus act like global variables. The functions even need to be global to let doctest find them without altering the user's code. (like the usual 
```
import doctest
doctest.testmod()
```
So after the run, we want to undef all functions, so that we only have access to what is currently defined in the script. So yes, functions defined by the user are removed from the globals, but not in the code string supplied by the user. So if the user edits it slightly to fix a mistake, the functions will be defined again in the next run.

chvp (approved):

PR #33, opened 2021-11-21
Add CodeMirror 6 editor

### LINKED ISSUE #2: Code editor
Evaluate different potential code editors on several features: flexibility, code support, usability, documentation, ...
Make a good selection and integrate it into the application.

- [x] Evaluate
- [x] Compare
- [x] Choose
- [ ] Integrate

### LINKED ISSUE #34: Keep code while switching language
Switching to another programming language clears the code editor. Keeping the code in the editor, but updating the syntax highlighting to the newly selected language seems like a better UX choice.

winniederidder:
This pull request replaces the text area to enter code with an instance of [CodeMirror 6](https://codemirror.net/6/).

codemirror-example

CM6 was chosen over alternatives such as Ace and Monaco because of its better documentation and extensibility.

image

For now, only basic features such as line numbers, auto-closing brackets and syntax highlighting are supported. It is possible to enable more advanced language features such as auto-complete and highlighting incorrect syntax, but they will be added in a later update as part of #46.

Closes #2 
Closes #34

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "webpack-dev-server": "3.11.1"
   },
   "dependencies": {
+    "@codemirror/autocomplete": "^0.19.8",
     "@codemirror/basic-setup": "^0.19.0",
     "@codemirror/lang-javascript": "^0.19.3",
     "@codemirror/lang-python": "^0.19.2",
--- a/package.json
+++ b/package.json
@@ -34,9 +34,15 @@
   },
   "dependencies": {
     "@codemirror/autocomplete": "^0.19.8",
-    "@codemirror/basic-setup": "^0.19.0",
+    "@codemirror/closebrackets": "^0.19.0",
+    "@codemirror/commands": "^0.19.5",
+    "@codemirror/comment": "^0.19.0",
+    "@codemirror/fold": "^0.19.1",
+    "@codemirror/history": "^0.19.0",
     "@codemirror/lang-javascript": "^0.19.3",
     "@codemirror/lang-python": "^0.19.2",
+    "@codemirror/rectangular-selection": "^0.19.1",
+    "@codemirror/search": "^0.19.3",
     "comlink": "^4.3.1"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -4,6 +4,8 @@
   "private": true,
   "homepage": ".",
   "devDependencies": {
+    "@codemirror/language": "^0.19.5",
+    "@types/codemirror": "^5.60.5",
     "@typescript-eslint/eslint-plugin": "^4.5.0",
     "@typescript-eslint/parser": "^4.5.0",
     "autoprefixer": "^10.4.0",
@@ -31,6 +33,9 @@
     "webpack-dev-server": "3.11.1"
   },
   "dependencies": {
+    "@codemirror/basic-setup": "^0.19.0",
+    "@codemirror/lang-javascript": "^0.19.3",
+    "@codemirror/lang-python": "^0.19.2",
     "comlink": "^4.3.1"
   },
   "scripts": {
--- a/src/CodeEditor.ts
+++ b/src/CodeEditor.ts
@@ -5,6 +5,7 @@ import { javascript } from "@codemirror/lang-javascript";
 import { ProgrammingLanguage } from "./ProgrammingLanguage";
 import { python } from "@codemirror/lang-python";
 import { LanguageSupport } from "@codemirror/language";
+import { autocompletion } from "@codemirror/autocomplete";
 
 function getLanguageSupport(language: ProgrammingLanguage): LanguageSupport {
     switch (language) {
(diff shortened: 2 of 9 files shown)


bmesuere (changes requested):
Functionally this seems to work. The editor is initialized with 10 lines and grows bigger if you add more lines. The output area also grows then the editor becomes bigger which should not be the case. If we properly layout the page, we will probably want to put the editor in fixed-height div with scrollbars (or configure codemirror so it behaves that way).

Can you add a bit more information to the PR start about which features are available in codemirror and which are enabled where? e.g. Syntax highlighting, auto complete, syntax errors, linting, ...

winniederidder:
We currently make use of  @autocomplete/basic-setup, which contains pretty much all things you listed (as can be seen [here](https://github.com/codemirror/basic-setup/blob/main/src/basic-setup.ts). The question is whether we want to keep it like that, or include all things mentioned here ourselves in our files. This would allow configuration when needed and might make it more obvious which features are present.

pdawyndt:
I would also like to second Bart's first paragraph: make sure editor takes up full browser height (except for some header/footer space) so that it doesn't need to grow/shrink as the text is edited. Initial code should not be 10 empty lines of code, but just empty code string. See [codeshare](https://codeshare.io/) using CodeMirror as an example.

The user settings they provide are: syntax highlighting (set programming language), tab size, theme and keymap (sublime, vim or emacs).

winniederidder:
Once this branch is merged, I can rebase the other open pull request, as this one has the same problem and depends on this branch.

package.json
```diff
@@ -4,6 +4,8 @@
   "private": true,
   "homepage": ".",
   "devDependencies": {
+    "@codemirror/language": "^0.19.5",
```
bmesuere, line 7: What does this package do? Is there a reason it's in the dev dependencies?

src/CodeEditor.ts
```diff
@@ -0,0 +1,64 @@
+import { EditorState, basicSetup } from "@codemirror/basic-setup";
+import { EditorView, keymap } from "@codemirror/view";
+import { indentWithTab } from "@codemirror/commands";
+import { javascript } from "@codemirror/lang-javascript";
+import { ProgrammingLanguage } from "./ProgrammingLanguage";
+import { python } from "@codemirror/lang-python";
+import { LanguageSupport } from "@codemirror/language";
+// import { autocompletion } from "@codemirror/autocomplete";
```
bmesuere, line 8: Why is this commented?

bmesuere (changes requested):
As asked previously, can you update the PR start with a list of features that is enabled in CodeMirror 6, possibly with some screenshots?

- Syntax highlighting is working for me
- In JS, I sometimes get snippet suggestions but pressing tab doesn't expand them (I think this is the shortcut in most editors?).
- In Python I wasn't able to trigger such suggestions
- In none of the language I was able to trigger autocomplete (e.g. functions of the std lib)
- In none of the languages, incorrect syntax was highlighted

package.json
winniederidder: This package contains the class LanguageSupport to provide typing on getLanguageSupport. There is a separe types folder in https://github.com/codemirror/language/blob/main/package.json but I am unable to just install those types using e.g. yarn add -D @types/codemirror-language or yarn add -D @types/codemirror/language.

src/CodeEditor.ts
winniederidder: I intend to use it when resolving https://github.com/dodona-edu/papyros/issues/46, so this is just to remove the unused identifier warning. Removing it would achieve the same goal but I intend to work on that issue in the near future.

winniederidder:
> As asked previously, can you update the PR start with a list of features that is enabled in CodeMirror 6, possibly with some screenshots?

I created a separate issue (#46) to handle this, as having this PR merged was more important to me so that other features I worked on could be reviewed.


> * Syntax highlighting is working for me
> * In JS, I sometimes get snippet suggestions but pressing tab doesn't expand them (I think this is the shortcut in most editors?).
The current keymap is the default (https://codemirror.net/6/docs/ref/#autocomplete.completionKeymap), which only uses enter. Vscode seems to support both tab and enter. 

> * In Python I wasn't able to trigger such suggestions
Indeed, JavaScript comes with a minimal basic amount of snippets, whereas Python has none out of the box.
> * In none of the language I was able to trigger autocomplete (e.g. functions of the std lib)
Indeed, the default behaviour is very minimal. The plan is to extend this in the future with our own CompletionSources.
> * In none of the languages, incorrect syntax was highlighted
This is probably again part of the default configuration. I will look into this when working on issue #46

bmesuere:
My question still stands. Can you update the PR start to be more specific about what is included in the PR and what is not.

The PR start should be the (up to date) summary of the changes.

bmesuere (approved):
I have updated the PR start as an example for future PRs.

Issue #43, opened 2021-11-25
I18N


winniederidder:
Currently, the page is based on a static webpage with hardcoded text in English. Instead, the elements should be added dynamically to the DOM, at which point we can replace the textual content by translated strings depending on the selection language.

PR #52, opened 2021-11-30
Allow configuring indent size

### LINKED ISSUE #37: Use 4 spaces per indentation level for Python
[PEP-8](https://www.python.org/dev/peps/pep-0008/) style guide for Python: Use 4 spaces per indentation level.

CodeMirror editor now replaces tabs with 2 spaces, and also adds 2 space indentation after header line.

winniederidder:
Depends on #43 
Closes #37 
By using a Compartment, changes to extensions in the CodeEditor can be done without recreating the entire editor. 

I used this approach for both the LanguageSupport and the indentUnit facets, allowing both to be updated more easily. This would allow us to let the user choose the amount of used spaces (with the default being 4).

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "@codemirror/lang-python": "^0.19.2",
     "@codemirror/rectangular-selection": "^0.19.1",
     "@codemirror/search": "^0.19.3",
+    "@codemirror/state": "^0.19.6",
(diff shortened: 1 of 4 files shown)


bmesuere (changes requested):
Can you rebase this one and set the PR target to main?

bmesuere (approved):

PR #54, opened 2021-12-01
Validate translations

### LINKED ISSUE #53: Test translations
When adding new features, it is easy to forget to add translations, even though the translate-function is called. A useful library for this is [i18n-extract](https://github.com/oliviertassinari/i18n-extract). It may be worthwhile to define a GitHub action to verify whether translations have been done correctly by using their API.

winniederidder:
This PR adds a JS script called ValidateTranslations that checks whether the provided translations are valid. A GitHub action is also defined to call this script on pushes to any branch.

The script is written in JavaScript as running it as a stand-alone TS file never fully worked. On the one hand, i18n-extract does not provide proper typing, requiring some hacks to get it into TypeScript (e.g. including the source code directly as it required some manual editing to even run). On the other hand, importing other modules did not work out of the box.
The JavaScript file can easily be run through Node, but the Translations.ts file had to be converted to JS for it to be imported. This isn't an impactful change, as they are simple objects where typing is quite irrelevant.

The current checks are: detecting any unused, missing, duplicated or dynamic keys, and failing if any of these are present (with failing only once all checks have been done to prevent rerunning often). Explictly empty keys (such as Papyros.ready) are considered as missing, but this is a special case so whitespace also does the trick.

Closes #53

### DIFF
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -29,6 +29,5 @@ jobs:
       - name: Remove deploy label
         uses: buildsville/add-remove-label@v1
         with:
-          token: ${{secrets.GITHUB_TOKEN}}
           label: deploy
           type: remove
--- a/.github/workflows/validate-translations.yaml
+++ b/.github/workflows/validate-translations.yaml
@@ -27,6 +27,7 @@ jobs:
       with:
         check_name: validate_translations
         linter_output_path: translationIssues.txt
+        regex: '^(?<filename>.*?):(?<lineNumber>\d+):(?<columnNumber>\d+): (?<errorCode>\w\d+) (?<errorDesc>[\s|\w|.]*)'
         commit_sha: ${{ github.sha }}
       env:
         GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
--- a/.github/workflows/validate-translations.yaml
+++ b/.github/workflows/validate-translations.yaml
@@ -5,11 +5,11 @@ on:
   workflow_dispatch:
 
 jobs:
-  javascript:
-    name: JavaScript
+  validate_translations:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - name: Checkout
+      uses: actions/checkout@v2
     - name: Set up Node
       uses: actions/setup-node@v2
       with:
@@ -21,6 +21,7 @@ jobs:
     - name: Check translations
       run: |
           yarn run validate:translations
+    - name: Add annotations
       uses: pytorch/add-annotations-github-action@master
       with:
         check_name: Check translations
--- a/.github/workflows/validate-translations.yaml
+++ b/.github/workflows/validate-translations.yaml
@@ -21,6 +21,7 @@ jobs:
     - name: Check translations
       run: |
           yarn run validate:translations
+          cat translationIssues.txt
     - name: Add annotations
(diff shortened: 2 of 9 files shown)


bmesuere (commented):
👍 Looks good. Do we also want to check for unused translations? Maybe in that case, we shouldn't error, but we can add them as annotations to the code, e.g. using https://github.com/pytorch/add-annotations-github-action

winniederidder:
> 👍 Looks good. Do we also want to check for unused translations? Maybe in that case, we shouldn't error, but we can add them as annotations to the code, e.g. using https://github.com/pytorch/add-annotations-github-action

Not entirely sure how these annotations would work. Unused translations would only be present in the Translations.js file. How would these be added?

Regarding my question to the issue with dynamic translation keys (with interpolated strings causing the static analysis to deem everything used), how do you think we best tackle that? I think it would be possible to make these explicit, which would make the analysis more useful.

bmesuere:
Regarding annotations: the action I linked lets you add github annotations (by writing filename, line number and message to a file). They then show up in the github UI. In this case, I would add such annotation to the lines containing unused translations.

Regarding dynamic translation keys: what is your question exactly? I probably missed it somewhere.

winniederidder:
> Regarding annotations: the action I linked lets you add github annotations (by writing filename, line number and message to a file). They then show up in the github UI. In this case, I would add such annotation to the lines containing unused translations.

I will take a look then, thanks!

> Regarding dynamic translation keys: what is your question exactly? I probably missed it somewhere.

I edited the original post so the question wasn't there anymore. In essence, if dynamic translation keys are used (i.e string interpolation in the translation key like `Papyros.${programmingLanguage}`) then this script becomes quite useless. It employs static analysis and cannot determine which values could be passed here, so it deems every key used. My question was thus whether we want to forbid dynamic keys. This might be a bit of a hassle for the programmer, but it does allow static analysis to be useful. I opted to indeed let go of dynamic keys and added the forbidDynamic check to the script to make this clear.

bmesuere:
Ok now I see. I think this is why they often have nested keys, e.g. papyros.programming_language.${language}, then only do interpolation in such subkeys.

winniederidder:
> Ok now I see. I think this is why they often have nested keys, e.g. papyros.programming_language.${language}, then only do interpolation in such subkeys.

That indeed sounds like it could work. Do you want me to integrate that, or keep this as this is and merge it?

bmesuere:
It seems cleaner to have nested translations to be able to group things, no?

winniederidder:
@bmesuere This branch should now also add annotations to files where a certain kind of translation issue was detected. It is thus ready for review and hopefully merging.

bmesuere (changes requested):
I wanted to test this (by adding a manual trigger to the GH Action), but it seems there is a syntax error in the workflow file. In addition, the name of the action looks funny in the GH Actions menu, so double check.

I would also temporarily add an extra translation key to check if the annotation shows up in the GitHub UI.

winniederidder:
@bmesuere Annotations from testing are available on https://github.com/dodona-edu/papyros/actions/runs/1545351976, but I am not exactly sure how to use them properly. Is this our expected output, or am I misinterpreting something?

winniederidder:
@bmesuere Annotations are now present in the "files changed" tab. Once you agree with the formatting, I can remove the testing keys and then this can be merged?

bmesuere (approved):
nice, looks good!

PR #45, opened 2021-11-27
Add I18N support

winniederidder:
This PR adds support for internationalization using [i18n-js](https://github.com/fnando/i18n-js).

dutch-papyros
This library was chosen due to ease of use and already being used in Dodona so this should help with the integration. Currently, English and Dutch translations are provided in src/Translations.ts.

In order to be able to translate the text, more control over the DOM was required. Instead of relying on an index.html file containing all the required elements, they are now created at run-time and filled in with the correct translation in the function renderPapyros. In order to provide this API, the Papyros.ts file was refactored to use a class-based approach, that now contains a static method fromElement to add it to an existing page. I tried reducing the scope pollution by combining related items into structured objects (such as the CodeState, InputState and StateManager) to hopefully aid in code readability and maintainability. This design will probably be fine-tuned later based on feedback and personal experience working with it.

Closes #43 
Closes #49 
Seems to have a positive impact on #44 , as the initial render is a lot less intense when reloading, so visual artefacts are reduced.
Perhaps this can be closed as well.

PR #55, opened 2021-12-02
Add stand alone mode

### LINKED ISSUE #50: Add option for stand alone mode
When constructing the application, an option should be passable that states whether the application is run in a stand alone mode. 
This option would change what is rendered (such as the programming language select, as this is fixed when openend within the scope of a Dodona exercise)

winniederidder:
This PR adds a config object to Papyros, with a new addition called standAlone. This boolean option tells Papyros to work in stand-alone mode, meaning it is its own application and not part of something bigger (such as Dodona).  This allows for conditional styling and rendering, as not every feature is relevant when not in stand-alone mode (such as a selector for the programming language, as this is most likely fixed and provided by the caller). The bar above the editor would then look like this:
stand-alone-papyros.

The current usage is disabling/enabling the header (containing the title and the language select) and the programming language select.
The stand-alone mode can thus now dynamically be translated by using urlParams, as follows.
stand-alone-papyros

This option could be used la

ter on to disable/enable other features or more agressive HTML-class changes if needed.

Depends on #45 
Closes #50 .

### DIFF
--- a/build/index.html
+++ b/build/index.html
@@ -13,7 +13,7 @@
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
   <link rel="manifest" href="manifest.json" />
-
+  <link href="https://cdn.jsdelivr.net/npm/@mdi/font@5.x/css/materialdesignicons.min.css" rel="stylesheet" type="text/css">
   <!-- Wait with running the script until the elements are loaded as they are needed immediately -->
   <script src="./bundle.js" defer></script>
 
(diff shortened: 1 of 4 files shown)


bmesuere (approved):
👍 I think this is a good time to start expanding the readme in the root of this repo to describe the available options. I'll make a separate issue for this.

I am a bit hesitant to add the mdi font here since it's huge. I know we already use it in Dodona, but I'm not happy about it there as well 😉

winniederidder:
> 👍 I think this is a good time to start expanding the readme in the root of this repo to describe the available options. I'll make a separate issue for this.
> 
> I am a bit hesitant to add the mdi font here since it's huge. I know we already use it in Dodona, but I'm not happy about it there as well 😉

Oh, I forgot to consider the size of MDI. This element however is only present in stand-alone mode, in which case the package size is less important I think? I could also just snatch a static svg or something simple to replace this icon, as I don't think I will be needing many.

PR #57, opened 2021-12-03
Fix callback return value

### LINKED ISSUE #27: Reading input is too slow
Reading input from stdin (`input()`) that has been preset in the input box is way too slow, even for a few lines of input.

### LINKED ISSUE #56: input() prints and returns None
Running this code:

```python
print(input() is None)
```

shows this output:

```
None
True
```

1. `input()` **prints** `None` directly, without help from `print()`, which is weird.
2. `input()` **returns** `None` regardless of the actual user input.

winniederidder:
This PR restores the functionality of input() in Python. The callback returned void, causing input to never receive a value and thus printing None. This is tied to the fact that I always print the value that the user gave. In a terminal, after calling input, the user is expected to type something and hit enter. This behaviour is emulated in the output window, by printing the prompt and then the obtained value. I noticed this causes some confusion, so this might only be desirable in an explicit REPL environment. The question is thus whether this extra print should also be removed as part of this PR.

Closes #56 
Can close #27  @pdawyndt ?

### DIFF
--- a/src/Backend.ts
+++ b/src/Backend.ts
@@ -36,7 +36,7 @@ function getInputCallback(inputTextArray?: Uint8Array, inputMetaData?: Int32Arra
 }
 
 export abstract class Backend {
-    onEvent: (e: PapyrosEvent) => void;
+    onEvent: (e: PapyrosEvent) => any;
     runId: number;
 
     constructor() {
--- a/src/Papyros.ts
+++ b/src/Papyros.ts
@@ -299,7 +299,6 @@ export class Papyros {
         } finally {
             const end = new Date().getTime();
             this.stateManager.setState(PapyrosState.Ready, t("Papyros.finished", { time: end - start }));
-            this.inputState.inputArea.value = "";
             this.inputState.lineNr = 0;
         }
     }
--- a/src/Papyros.ts
+++ b/src/Papyros.ts
@@ -194,6 +194,7 @@ export class Papyros {
             stopBackend(this.codeState.backend);
             this.codeState.programmingLanguage = programmingLanguage;
             this.codeState.editor.setLanguage(programmingLanguage);
+            this.codeState.outputArea.value = "";
             await this.startBackend();
         }
     }
--- a/src/workers/python/PythonWorker.worker.ts
+++ b/src/workers/python/PythonWorker.worker.ts
@@ -39,7 +39,7 @@ class PythonWorker extends Backend {
         // Python calls our function with a dict, which must be converted to a PapyrosEvent
         const eventCallback = (data: any): void => {
             const jsEvent: PapyrosEvent = "toJs" in data ? data.toJs() : Object.fromEntries(data);
-            this.onEvent(jsEvent);
+            return this.onEvent(jsEvent);
         };
         this.pyodide.globals.get("__override_builtins")(eventCallback);
         this.globals = new Map((this.pyodide.globals as any).toJs());


pdawyndt:
Input disappears from stdin field after it has been processed. Is this on purpose. For me, it was more useful to keep the input in place in the stdin field.

winniederidder:
> Input disappears from stdin field after it has been processed. Is this on purpose. For me, it was more useful to keep the input in place in the stdin field.

At one of the weekly meetings, it was mentioned that this should be reset after every run, so I took it away. This is no longer the case then?

bmesuere:
> > Input disappears from stdin field after it has been processed. Is this on purpose. For me, it was more useful to keep the input in place in the stdin field.
> 
> At one of the weekly meetings, it was mentioned that this should be reset after every run, so I took it away. This is no longer the case then?

We should decide between one of two modes:
- If we use synchronous input, this should be cleared
- If we ask to enter input before running the code, it should stay

The combination of both is very confusing. Entering it interactively is terminal-like, but rerunning (without clearing) suddenly reuses your old history.

winniederidder:
> An alternative solution, which might even we better if we stick with sync input, is combining the input and output fields as is suggested in #56

In #4 some more issues related to synchronous input were brought to light, so we should consider those extra risks in synchronous IO mode. In order to prevent confusion, it is probably best to provide this as an option to the user, so that Papyros either runs in a sort of REPL mode (which could be extended to become a proper REPL) using only synchronous IO, or in a mode where all input is provided up front.

alexmojaki:
> In #4 some more issues related to synchronous input were brought to light, so we should consider those extra risks in synchronous IO mode.

I think you've solved synchronous input quite nicely?

I think that having a proper REPL is an important feature that makes a big difference to the learning experience. It allows quick, frictionless experimentation and exploration so that users don't hesitate to ask questions and then answer them for themselves. So I think providing a REPL is more important than handling `input()` in the most optimal way possible. And if you have the UI for a REPL, then handling `input()` within that REPL (i.e. integrated with the output, unlike the current UI) is much easier.

BTW, if you still have concerns about synchronous input, note that a REPL doesn't need to use that stuff (except ofc when running code that uses stdin).

alexmojaki:
I've only just looked at Dodona itself now, and that significantly changes things. In futurecoder, `input()` is only introduced quite late, and it's used specifically for interactive programs like the tic-tac-toe game. Other (i.e. most) exercises essentially require defining a function that simply receives arguments of the correct type. Since many (all?) Dodona exercises are based on stdin, and sometimes the input consists of many lines, having a separate input area that doesn't clear with each run is obviously much more important for you than interactive terminal-like input. In fact, showing that input again in the output section will probably feel like annoying noise for most exercises, especially exercises like "Sum of N numbers".

winniederidder:
It indeed might look noisy. However, in this image that I posted in a similar thread, I think it is useful.
input-papyros
I think a good compromise might be to only print if the prompt value is also not empty, so that it remains clear which value was used for each input call. This might be useful when a student is debugging

alexmojaki:
These are some good points, but:

- Showing the input for some programs and not others will likely confuse students.
- Showing some inputs (when a prompt is given) but not others in the same program will also cause confusion
- Some students may write `print("Give a number:"); x = int(input())`

Can you link to any exercises in Dodona that ask for an 'interactive' program that includes prompts?

pdawyndt:
@alexmojaki Here's the [`sum of two integers`](https://dodona.ugent.be/en/activities/1555355466/) exercise that we use as the first exercise. Actually, the Python judge that evaluates submissions allows to ignore prompts from the built-in function `input` (which is what I do for all exercises in my course), such that students have complete freedom to use any prompts they want or not prompts at all.

alexmojaki:
I think there's a misunderstanding somewhere. That linked exercise doesn't mention prompts at all, so I wouldn't expect a student to use them while practicing in the scratchpad. I'm assuming you were thinking of https://dodona.ugent.be/en/courses/290/series/2518/activities/444603041/ .

In any case the program is not really 'interactive', it's just using stdin as a mechanism to receive input values. I'm asking if there are exercises where prompts have clear value in the program, where receiving input is interleaved with other output or computation so a user needs to be told what the program is asking for. For example see the tic-tac-toe game: the very last link in https://futurecoder.io/course/#toc. In particular the 'judge' shows the following expected output when an incorrect program is submitted:

```
  1 2
1  |
  -+-
2  |
A to play:
<input: 1>
<input: 1>
  1 2
1 A|
  -+-
2  |
B to play:
<input: 1>
<input: 2>
  1 2
1 A|B
  -+-
2  |
A to play:
<input: 2>
<input: 1>
  1 2
1 A|B
  -+-
2 A|
A wins!
```

That's an example where the 'prompts' and the user input need to be shown in the output for the user to understand.

All the exercises I'm seeing on Dodona have a clear division between handling input at the start of the program and then doing everything but input in the rest of the program. In that case:

- Entering input in a separate area once makes much more sense than an interactive terminal-like input method
- Students have very little reason or incentive to use prompts at all, it basically amounts to print debugging
- Showing the input again in the output area is not helpful

pdawyndt:
Input (prompts) and output may differ between implementations (e.g. print results as soon as they are computed vs keeping all results in a data structure and only print at the very end of the code). To give students freedom between these alternatives, the Python judge allows to ignore prompts, so that only the expected output (not from prompts) is compared with what is expected (without prompts as well). The judge can  also take into account the prompts, but than they must be in sync between the expected and generated output (input never is mixed in the output).

However, when running code while developing, prompts might help students to see what data is expected. For this case, @winniederidder wants to provide the option to set input a priori (and mimick as it is entered interactively) or enter it interactively when the code is running in the same "interactive" window where the output appears. However, the interactive editing is not provided yet, so this is still work in progress.

For me, running the code and testing the code are two separate worlds.

alexmojaki:
> For me, running the code and testing the code are two separate worlds.

I don't think it will feel like that for students. If the scratchpad and the submission testing behave differently that will be surprising and confusing. Besides, they will not always develop a perfect solution in the scratchpad and then pass the tests the first time they submit. Development will often happen in the 'official' area as students will constantly think they just have to change one more thing to fix their program.

Since the judge ignores prompts and they are not shown in stdout in the tests, then it seems the output area of the scratchpad must display prompts in a different colour from other output, maybe the same colour as the input.

Longer term, if there is a configuration option to choose between interactive and preset input, I think ideally there should also be an option to hide the preset input in the output area. This would also hide the prompts to emphasise to the user that they are not part of the output seen by the judge.

bmesuere (approved):

alexmojaki:
I see that both prompts and input are not shown in the output regardless of mode. Is that a result of the discussion here? I was a little surprised to not see prompts in batch mode but it might be for the best.

PR #64, opened 2021-12-03
Show terminate button only when appropriate

### LINKED ISSUE #61: Terminate button activated upon page load
When the Papyros page is loaded, the `Terminate` button is activated as long as the Python runtime hasn't loaded. This should not be the case (user did not run any Python code that could be terminated yet). Only the `Run` button should be disabled until the Python runtime becomes available.

winniederidder:
This PR adds finer control over when the terminate button is visible. The original code only compared to Ready, whereas Loading is also a state that should be preserved.

As a bonus, I added logs and  early returns when runCode/terminate are called from states when they shouldn't be callable. This should only occur when a user tampers with the DOM, but might also protect against coding errors.

Closes #61

### DIFF
--- a/src/Papyros.ts
+++ b/src/Papyros.ts
@@ -82,7 +82,7 @@ function renderPapyros(parent: HTMLElement, programmingLanguage: ProgrammingLang
         </button>
         <button id="terminate-btn" type="button" 
             class="text-white bg-red-500 border-2 m-3 px-4 inset-y-2 rounded-lg
-            disabled:opacity-50 disabled:cursor-wait" hidden>
+            disabled:opacity-50 disabled:cursor-wait">
             ${t("Papyros.terminate")}
         </button>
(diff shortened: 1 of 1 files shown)


pdawyndt:
Better UI if both the `Run` and `Terminate` buttons always stay in place, but are enabled/disabled when appropriate.

bmesuere:
> Better UI if both the `Run` and `Terminate` buttons always stay in place, but are enabled/disabled when appropriate.

I do not agree :) Both buttons can never be active at the same time and showing them both is a bit prominent (also, the "executing" cursor is shown over the stop button when nothing is running. I would change it back to a single button or make the stop button a lot less prominent. An alternative would be to show terminate as a link in the "status" message. We can leave it as is for now and tweak it later with the rest of the layout.

bmesuere (approved):

PR #69, opened 2021-12-06
Add input modes

### LINKED ISSUE #19: Strip quotes from stdout in JavaScript
Running this code to test output generation in JavaScript

```javascript
for (let name of ['Alice', 'Bob', 'Claire']) {
    console.log(`Hello, ${name}!`);
}
```

generates output containing double quotes:

```
"Hello, Alice!"
"Hello, Bob!"
"Hello, Claire!"
```

### LINKED ISSUE #68: Toggle between batch and single-line input mode
The current layout for handling input is a bit confusing. A toggle between batch mode (giving multiple inputs beforehand) and single-line (a single input area that is cleared upon pressing enter) would help making it clear when and where to enter data for the program.

### LINKED ISSUE #74: Disable enter button when needed
If the input field is disabled, the `Enter` button can still be clicked. The latter enables the input field and upon pressing it seems as if the code starts executing an infinite loop. All this would be avoided if the Enter button is enabled/disabled along with the input field.

winniederidder:
This PR adds input modes to Papyros. Currently, a select is used to choose between Interactive and Batch mode (represented by the InputMode enum). The input handling functionality has been moved to a separate class called InputManager to deal with (re)rendering of the input area.

papyros-input

Interactive mode is the default, it provides a single line input field with a button which acts like pressing enter.
Batch mode is the same as before, a multiline textarea that is not cleared.

The input area has a default placeholder that is replaced by the prompt if possible.

Closes #68 
Closes #19 as I had to edit that code anyway to improve the prompts.
Closes #74


pdawyndt:
Possible naming:

**English**

1. on demand, interactive, just-in-time, synchrone
2. preset, asynchrone

**Dutch**

3. op aanvraag, interactief, synchroon
4. vooraf ingesteld, asynchroon

pdawyndt:
Optimal "hybrid" approach still seems the possibility to enter answers interactively in the same edit panel where the prompt appears (with a blinking cursor indicating that input is expected), after lines from a preset input have been consumed first. UI-wise this seems the most challenging, but CodeMirror seems to provide some support for this.

winniederidder:
@bmesuere Following your remarks, there would not be a third/hybrid mode? Should batch mode throw an error when no more input is found, instead of waiting and accepting enter-presses?

bmesuere:
No, I would focus the text area in that case and display an additional message

winniederidder:
How would this work together with the link you mentioned to switch modes? Or is this by default active in Batch mode then?

winniederidder:
Current visuals for in interactive mode:
papyros-input
 Generically making only a part of the switching text a link is not very straightforward, so currently all of it functions as a clickable toggle.

bmesuere:
> How would this work together with the link you mentioned to switch modes? Or is this by default active in Batch mode then?

Yes, I think the way you implemented it right now is ok, except for maybe an additional help message if still waiting in batch mode.

winniederidder:
> Yes, I think the way you implemented it right now is ok, except for maybe an additional help message if still waiting in batch mode.

I'm thinking of making it a tooltip when they hover of the input area, or should I add a small question mark icon that they can hover over for help, or even something else?

pdawyndt:
Already looks pretty good. Some suggestions:

- Couldn't you put short instructions like "Enter your code here." or "Enter your input here." as default text in the editor and the input, to make clear what's expected from the user (disappears as soon as user starts editing)
- When switching to interactive mode, the edit field already contains some spaces. Would be easier if these weren't there.
- Is the flow clear if users switch between input and interactive mode while running code? Maybe we should disable switching between both modes while code is running.
- When switching away from and back to interactive mode, the previous batch input has disappeared. I would keep the input upon switching, but only use it when in interactive mode.
- Do we want to keep the interactivity of "batch mode" as is. Now, if the input of the batch mode is exhausted and more input is requested, you can still type additional input in the input field. Maybe it's more predictable if the batch input field is set in readonly mode while running code, and when all input is consumed and more input is requested, we could automatically switch to interactive mode as a sign that more input is requested by the execution. 
- We could even think about highlighting the line of input that will be read next in the readonly input field (might be useful later when we introduce debugging).

bmesuere:
- Placeholder text doesn't work since we give the input field automatic focus if waiting waiting for input
- I wouldn't add default text to the code editor itself. You will have to remove this manually every time you open the editor with only a small first time benefit
- I also noticed the extra spaces
- Agreed with disable switching if running, but I wouldn't make this a priority
- Batch input should indeed stay in the text area when switching to interactive mode and back
- I would *not* switch to interactive mode when the batch input is exhausted. The current behaviour with an info message telling to add input seems clear to me and the expected thing to do. The additional lines will be part of the input on next run where this wouldn't be the case if you actually switched modes.
- I wouldn't bother with highlighting lines for now. If not paused for debugging, this will almost never be visible or flash very rapidly.

winniederidder:
> * Placeholder text doesn't work since we give the input field automatic focus if waiting waiting for input

Even when focused, the placeholder is still there if the field is empty. Or is this a browser specific thing? It works just fine for me.

> * I wouldn't add default text to the code editor itself. You will have to remove this manually every time you open the editor with only a small first time benefit

There is no default text that needs removing. The code editor just also has a placeholder that is cleared when code is entered.

> * I also noticed the extra spaces

This was a formatting issue when using backticks in JavaScript and has been resolved.

> * Agreed with disable switching if running, but I wouldn't make this a priority

Currently, I added an automatic switch to interactive mode when there is no more input in batch mode, combined with @pdawyndt 's suggestion to keep track of the previous batch input. If such a switch occurs during a run, the mode is also restored at the end of the run.

> * Batch input should indeed stay in the text area when switching to interactive mode and back

This has been implemented.

> * I would _not_ switch to interactive mode when the batch input is exhausted. The current behaviour with an info message telling to add input seems clear to me and the expected thing to do. The additional lines will be part of the input on next run where this wouldn't be the case if you actually switched modes.

I implemented that, so you can tyr it out. If you still dislike it, then I can revert it.

src/Translations.js
```diff
@@ -3,12 +3,19 @@
...
+    "code": "Code",
+    "code_placeholder": "Enter your code here",
+    "input": "Input",
+    "input_placeholder": {
+        "interactive": "Provide input here. Press enter to send.",
+        "batch": "Provide all input required by your code here.\n" +
+            "You can enter multiple lines by pressing enter."
+    },
+    "output": "Output",
+    "output_placeholder": "The output of your code will be shown here",
```
bmesuere, line 15:
```suggestion
    "output_placeholder": "The output of your code will appear here",
```

src/Translations.js
```diff
@@ -3,12 +3,19 @@
 
 const ENGLISH_TRANSLATION = {
     "Papyros": "Papyros",
-    "enter_code": "Enter your code here",
-    "enter_input": "Provide input for your code here",
-    "code_output": "Code output",
+    "code": "Code",
+    "code_placeholder": "Enter your code here",
```
bmesuere, line 7:
```suggestion
    "code_placeholder": "Write your code here and click run to execute...",
```

If possible, make this "JavaScript code" or "Python code" depending on the mode.

src/Translations.js
```diff
@@ -3,12 +3,19 @@
...
+    "input": "Input",
+    "input_placeholder": {
+        "interactive": "Provide input here. Press enter to send.",
```
bmesuere, line 10:
```suggestion
        "interactive": "Provide input and press enter to send",
```

src/Translations.js
```diff
@@ -25,13 +32,25 @@ const ENGLISH_TRANSLATION = {
...
+    "enter": "Enter"
 };
 
 const DUTCH_TRANSLATION = {
     "Papyros": "Papyros",
-    "enter_code": "Voer hier je code in",
-    "enter_input": "Geef hier de invoer voor je code in",
-    "code_output": "Resultaat van de code",
+    "code": "Code",
+    "code_placeholder": "Voer hier je code in",
```
bmesuere, line 45:
```suggestion
    "code_placeholder": "Schrijf hier je code en klik op 'voer uit' om uit te voeren...",
```
Still not totally happy with this phrasing but can't come up with a synonym for "uitvoeren"

bmesuere (changes requested):
Looks good. My main remarks are:
- I don't like the automatic switch from batch to interactive when the input is exhausted. This goes against an explicit choice made by the user (to use batch mode) and doesn't explain why it happens. I prefer staying in batch mode, highlighting the input area and adding a message at the bottom stating "Predefined input was exhausted. Waiting for additional input".
- Placeholders: they only disappear when something is entered, not when they have the focus (sorry, my bad). Current behaviour is ok, but a added some suggestions regarding the wording. I would not use the placeholder in the batch input mode because it is less common for text areas and since the input isn't cleared after a run it isn't always visible. Instead I would add a small help text at the bottom. Something like "Lines entered here will be passed as input to your code"

Smaller suggestions:
- Give the CM6 editor focus when loading the page
- Add a tooltip (title attribute) to the interactive input when it's disabled to explain why it is disabled.
- I would experiment with moving the run button to the bottom of the code area (left aligned) and add the status message on the same line below the code area, but right aligned.
- The placeholder (and text) sticks to the border of the input and text area. Add some padding.

pdawyndt:
Maybe add status messages to status bar at the bottom of the CM6 editor, as in this [example from CM6](https://codemirror.net/6/examples/panel/) itself (see bottom of page).

image

src/Translations.js
```diff
@@ -25,13 +32,25 @@ const ENGLISH_TRANSLATION = {
...
+    "input": "Invoer",
+    "input_placeholder": {
+        "interactive": "Geef hier je invoer in. Duw op enter om te verzenden.",
```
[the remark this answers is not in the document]
pdawyndt: What about using the actual prompt from `input('Question: ')` here (with a default text as fallback if no prompt (empty string) is given, instead of putting the prompt in the output panel? This way, the right question is asked, and the output panel is not cluttered with the prompts.

src/Translations.js
pdawyndt: One option I see is naming the two buttons `Run` and `Stop` (same in Dutch). Then the sentence makes more sense in Dutch. We could also add the typical triangle and square icons to the run and stop buttons, as an extra visual aid.

Maybe also drop the word "here" for consistency with the input field. The word "je" may also be dropped.

pdawyndt:
Use same font in "Output" and "Input" panels as in code editor.

bmesuere:
Dodona uses this font stack for code: `Monaco, Menlo, "Ubuntu Mono", Consolas, source-code-pro, monospace`

bmesuere:
@winniederidder there seems to be a type issue while deploying

winniederidder:
Managed to reproduce it by removing node_modules, looking into the dependency issue now.

winniederidder:
@bmesuere I openend an issue over on [CodeMirror](https://github.com/codemirror/CodeMirror/issues/6846) as I believe this is due to a recent new version of @codemirror/state. This explains why I needed to reinstall node_modules before encountering this issue myself.

pdawyndt (approved):

bmesuere (changes requested):
For me (on chrome/mac) the output text area still uses the "normal" font instead of the code font. The rest of the functionality looks good :+1:

I will now start reviewing the code.

package.json
```diff
@@ -5,13 +5,10 @@
   "homepage": ".",
   "devDependencies": {
     "@babel/traverse": "^7.16.3",
-    "@codemirror/language": "^0.19.5",
-    "@types/codemirror": "^5.60.5",
     "@types/i18n-js": "^3.8.2",
     "@typescript-eslint/eslint-plugin": "^4.5.0",
     "@typescript-eslint/parser": "^4.5.0",
     "autoprefixer": "^10.4.0",
-    "babel-eslint": "^10.1.0",
```
bmesuere, line 14: why is this no longer needed?

bmesuere (commented):
After we get 0.1 out of the door, we should probably focus on consolidating the code and restructuring things a bit. For example, the InputManager class now contains logic to pass around input using the buffers, the logic to switch between input modes, the code to render the input and even some css. I won't open the discussion now, but this doesn't seem very manageable if we further extend the functionalities op Papyros.

pdawyndt:
I also have the default font (Windows 10, Chrome).

I'm still in favor of switching to CodeMirror for the output panel. Features like line numbers, no-wrapping are also useful for this panel, and probably also some of the other features in future iterations.

package.json
winniederidder: Babel-eslint is deprecated in favor of @babel/eslint-parser. I'm not entirely sure to what extent babel-eslint itself was being used, but I just noticed that message during the build-fixing process so I opted to remove it.

winniederidder:
> After we get 0.1 out of the door, we should probably focus on consolidating the code and restructuring things a bit. For example, the InputManager class now contains logic to pass around input using the buffers, the logic to switch between input modes, the code to render the input and even some css. I won't open the discussion now, but this doesn't seem very manageable if we further extend the functionalities op Papyros.

By using Tailwind, the css is inlined in the classes, so any rendering code in this application will contain css in that sense.
As you describe it, the component takes care of sending input, handling the input mode and rendering, which sounds similar to the React approach I'm used to. Could you open a separate issue on this to elaborate further then on what the downsides are of this approach?

bmesuere:
I was talking about the general stuff like the font stack that's in a variable, not the classes. Unless I'm missing something this could go in our css file.

The other stuff is bad separation of concerns. If you go over the input manager class, it's full of "if mode is this, do that; otherwise do something else" this is a clear code smell that the 2 input modes should be separate components (probably with a common super class ("InputProvider"?)). Its common interface should be to return the next input line. Another class should then "manage" these things, request lines from the right component and pass them to the worker.

winniederidder:
Regarding button icons, the usage of MDI icons was discouraged. Which source can I use then @bmesuere ? This is to add the usual play/stop icons to the relevant buttons.

winniederidder:
I created a new issue for the rework and some other elements to make sure they are remembered. (#76 )

bmesuere:
> Regarding button icons, the usage of MDI icons was discouraged. Which source can I use then @bmesuere ? This is to add the usual play/stop icons to the relevant buttons.

I would just inline the svg here

bmesuere (approved):

PR #77, opened 2021-12-12
Enable releasing on NPM

### LINKED ISSUE #59: Add readme
The readme in the root of the repo can be expanded with:
- the goal of the project
- a link to the currently deployed version
- minimal instructions to use this yourself
- a list of parameters available at initialisation time

### LINKED ISSUE #60: Release as npm package
To be able to use this in other projects (such as Dodona), we should also release this as an npm package.
Contact @rien for more information on how to publish under the @dodona prefix and org. And how to build this using a github action.

You should pay attention to
- which module system is used to publish the package
- not including unneeded files
- including the type definitions

winniederidder:
This PR adds the necessary changes to build Papyros into a usable NPM module.
A README with usage instructions was added to provide general informations and a small guide on how to use the package.

The workers are indicated via the .worker.ts extension and are inlined by using worker-loader, preventing resolving errors.
An action was included to publish tagged branches to NPM. 

Closes #59 
Closes #60


winniederidder:
The following Webpack5 issue appears to be similar to our issue: https://github.com/webpack/webpack/issues/12719.

winniederidder:
Another related [issue](https://github.com/webpack/webpack/issues/7353) mentions skipping the bundling entirely and just providing the files, or using url-loader. A final alternative would be to just provide a script in the installation instructions to copy the workers to the public directory, as mentioned earlier.

winniederidder:
@bmesuere I found a fix by using worker-loader in combination with webpack 5 (even though [the docs](https://webpack.js.org/guides/web-workers/) mention that you don't need to use it) that I believe inlines the workers so no more resolution errors occur.  In a test project, I managed to include and use Papyros and its workers. The release flow is thus simply building, packing and publishing.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,30 @@
+name: Publish on NPM
+
+on:
+  push:
+    tags:
+      - "v*"
```
bmesuere, line 6: Why did you choose to publish on tag push and not on github release creation?

README.md
```diff
@@ -1,7 +1,81 @@
...
+  </a>
+</p>
 
-Python support is powered by Pyodide. \
-JavaScript is supported natively in the browser.\
-Other languages are currently out of scope, but the application is designed to be able to be extended.\
+Papyros is a programming scratchpad in the browser. It allows running arbitrary
+in various programming languages directly in your browser, no installation required.
+By taking away obstacles between students and coding, the learning experience becomes
+smoother and less error-prone.
```
bmesuere, line 18:
```suggestion
Papyros is a programming scratchpad in the browser. It allows running code
directly in your browser, no installation required. Right now, the focus in on providing a great experience for Python, while also supporting JavaScript.
By taking away obstacles between students and coding, the learning experience becomes
smoother and less error-prone. 
```

README.md
```diff
@@ -1,7 +1,81 @@
...
-Other languages are currently out of scope, but the application is designed to be able to be extended.\
+Papyros is a programming scratchpad in the browser. It allows running arbitrary
+in various programming languages directly in your browser, no installation required.
+By taking away obstacles between students and coding, the learning experience becomes
+smoother and less error-prone. 
+
+Papyros aims to be:
+
+- **Easy to use** by having minimal installation instructions and an intuitive user interface
+- **Flexible** to support many programming languages
```
bmesuere, line 23: I would leave this out. Since we only support 2 languages, the "many" is not really correct. The installation instructions from the first point is also invalid since you don't have to install a thing.

README.md
```diff
@@ -1,7 +1,81 @@
...
+Currently, Papyros provides support for the following programming languages:
+- Python, powered by [Pyodide](https://pyodide.org/en/stable/)
+- JavaScript, powered by your browser
+
+## Installation
+
+You can install Papyros on your system using npm:
+```shell
+npm install -g @dodona/papyros
+```
```
bmesuere, line 34: This should be replace by the use it online section. The "installation" section should be moved to the next "usage" section which should be renamed to "Using Papyros in your own project" or something like that.

src/library.ts
```diff
@@ -0,0 +1,9 @@
+export * from "./Constants";
```
bmesuere, line 1: Doesn't this add all these classes to the global scope? If so, we should reconsider this when refactoring later.

bmesuere (changes requested):

src/library.ts
winniederidder: Indeed, the choices made here will determine what other users of the package will be able to use.  Some things like the CodeEditor might also be useful for end users, so this is up for discussion.

bmesuere: You can still make thing available to users without clobbering the global scope by scoping them inside a papyros class or object.

public/inputServiceWorker.js
```diff
@@ -8,50 +8,56 @@ function sleep(ms) {
...
+                    });
+                    return moddedResponse;
+                });
+        }
     } else { // requests to other domains are unaltered
         promiseChain = fetch(event.request);
     }
     event.respondWith(promiseChain);
 });
+// Prevent needing to reload page to have working input
```
alexmojaki, line 57: Does this mean the cross origin isolation headers will sometimes be provided by the server instead of the service worker?

winniederidder: This is to allow the input url to work, in the case where there are no SharedArrayBuffers. Ideally, the server always provides the headers because a reload is required if the headers are not yet there upon page load (as the service worker was not active yet and could not add them), but I cannot do this without harming the user experience.

src/Papyros.ts
```diff
@@ -117,44 +58,60 @@ class PapyrosStateManager {
...
+                programmingLanguage,
                 t("Papyros.code_placeholder", { programmingLanguage })),
             backend: {} as Remote<Backend>,
-            outputArea: document.getElementById(OUTPUT_TA_ID) as HTMLInputElement,
             runId: 0
         };
+        const statusPanel = new StatusPanel();
         this.stateManager = new PapyrosStateManager(statusPanel);
         this.inputManager = new InputManager(() => this.stateManager.setState(PapyrosState.Running), inputMode);
+        this.inputURL = location.host;
```
alexmojaki, line 114: Is this still used?

src/Papyros.ts
```diff
@@ -215,9 +177,45 @@ export class Papyros {
...
+            window.localStorage.removeItem(RELOAD_STORAGE_KEY);
+            return true;
+        } else {
+            if (typeof SharedArrayBuffer === "undefined") {
+                papyrosLog(LogType.Important, "SharedArrayBuffers are not available. ");
+                if (!serviceWorkerRoot || !serviceWorkerName || !("serviceWorker" in navigator)) {
+                    papyrosLog(LogType.Important, "Unable to register service worker. Please specify all required parameters and ensure service workers are supported.");
+                    return false;
+                }
+                if ("serviceWorker" in window.navigator) {
```
alexmojaki, line 200: You can remove this condition since you checked it above.

src/InputManager.ts
```diff
@@ -21,39 +21,24 @@ interface InputSession {
...
-                new SharedArrayBuffer(Uint8Array.BYTES_PER_ELEMENT * 1024));
-            // 2 Int32s:
-            // index 0 indicates whether data is written
-            // index 1 denotes length of the string
-            this.inputMetaData = new Int32Array(
-                new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT * 2));
-        } else {
-            papyrosLog(LogType.Important, "Using serviceWorker for input");
-        }
+        this.channel = makeChannel({ atomics: {} })!; // by default use Atomics
```
alexmojaki, line 38:
```suggestion
        this.channel = makeChannel()!; // by default use Atomics
```

Atomics is always used if it's available. The `atomics/serviceWorker` key is just to separate options into groups, not to specify what to use.

winniederidder: I wanted to make it more clear what was being used, but the comment also achieves that I suppose.

src/Papyros.ts
winniederidder: Indeed, slight oversight in the merge.

src/Papyros.ts
winniederidder: The channel takes care of that now, so indeed no longer used.

src/BackendManager.ts
```diff
@@ -1,23 +1,23 @@
-/* eslint-disable max-len */
+/* eslint-disable */
```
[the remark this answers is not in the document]
winniederidder: No longer needed as the way workers are imported changed: worker-loader doesn't cause eslint problems anymore.

bmesuere (approved):

PR #80, opened 2021-12-15
Better error handling

### LINKED ISSUE #1: Error handling
This issue consists of on the one hand properly displaying errors to the user while hiding the internals of the application (such as the web assembly parts) by filtering stack traces, and on the other hand catching anything that could go wrong. A list of example code files that should be handled are found in https://github.com/dodona-edu/universal-judge/issues/19. 

- [ ] Hide internals (wasm, unknown file, ...)
- [ ] Display errors in an expected fashion
- [ ] Handle common errors properly

### LINKED ISSUE #83: Only text gets printed in JavaScript
Only text gets printed in JavaScript:

image

A spurious empty lines gets printed in between two text outputs in JavaScript

image

winniederidder:
This PR adds functionality to provide better error handling to help the user with debugging.

To achieve this, the Python code was refactored to use a class-based approach to keep things modular.

Additionally, in stand-alone mode, examples are now provided per programming language for both demonstration and testing purposes. These can be used to show example error outputs, show the user typical programming constructs and help in debugging the application by having relevant code a mouse-click away.
PapyrosEvents now supprt a content property to distinguish between text, images, ...

For Python, we can use the friendly_traceback library (suggested by @alexmojaki) which can provide detailed information that can help users debug. An example output is show in the attached image. 
example-error

Errors are represented as follows:
```typescript
export interface FriendlyError {
    name: string;
    traceback?: string;
    info?: string;
    why?: string;
    what?: string;
    where?: string;
}
```
These components should, when filled in properly, allow us to provide detailed feedback to the user. 
The name is simply the name of the Exception class/object.
The traceback could be shown to provide more information to advanced users. By default, we limit this traceback to files under the user's control.
The info-value is displayed when hovering over the blue question mark, it gives generic information about the error type.
The why-value is a short description of what went wrong in the operation.
What is a short description of the error.
Where describes where in the source file this occurred.

Current issues: 
Design: how do we want to visualize each part?
Fallbacks: What if one or more parts are not present? 
I18n: friendly_traceback does not (yet) support Dutch translations. However, this language is important for us. Should I add translations to their repository? (Only useful if we are satisfied with the library).
Closes #1 
Closes #83

### DIFF
--- a/src/workers/python/PythonWorker.worker.ts
+++ b/src/workers/python/PythonWorker.worker.ts
@@ -70,10 +70,13 @@ class PythonWorker extends Backend {
             // run the code, potentially polluting the namespace
             // Functions and variables defined by the user become global
             // We need them to be global to let doctest work out of the box
-            const result = this.pyodide.runPython(code);
-            // Cleanup the scope after computations are done
-            this._cleanScope();
-            return result;
+            try {
+                return this.pyodide.runPython(code);
+            } finally {
+                // Cleanup the scope after computations are done
+                // Even in case of errors
+                this._cleanScope();
+            }
         } else {
             return this.pyodide.runPythonAsync(code);
         }
--- a/src/workers/python/PythonWorker.worker.ts
+++ b/src/workers/python/PythonWorker.worker.ts
@@ -30,11 +30,10 @@ class PythonWorker extends Backend {
     override async launch(onEvent: (e: PapyrosEvent) => void,
         inputTextArray?: Uint8Array, inputMetaData?: Int32Array): Promise<void> {
         await super.launch(onEvent, inputTextArray, inputMetaData);
-        const pyodide = await loadPyodide({
+        this.pyodide = await loadPyodide({
             indexURL: "https://cdn.jsdelivr.net/pyodide/v0.18.1/full/",
             fullStdLib: false
         });
-        this.pyodide = pyodide;
         await this.runCode(INITIALIZATION_CODE, 0);
         // Python calls our function with a dict, which must be converted to a PapyrosEvent
         const eventCallback = (data: any): void => {
@@ -71,7 +70,7 @@ class PythonWorker extends Backend {
             // Functions and variables defined by the user become global
             // We need them to be global to let doctest work out of the box
             try {
-                return this.pyodide.globals.get("__run_code")(code);
+                return await this.pyodide.globals.get("__run_code")(code);
             } finally {
                 // Cleanup the scope after computations are done
                 // Even in case of errors
--- a/src/workers/python/PythonWorker.worker.ts
+++ b/src/workers/python/PythonWorker.worker.ts
@@ -1,7 +1,7 @@
 import { expose } from "comlink";
 import { Backend } from "../../Backend";
 import { PapyrosEvent } from "../../PapyrosEvent";
-import { INITIALIZATION_CODE, INITIALIZE_PYTHON_BACKEND, RUN_PYTHON_CODE } from "./init.py";
+import { INITIALIZATION_CODE, INITIALIZE_PYTHON_BACKEND, PROCESS_PYTHON_CODE } from "./init.py";
 
 interface Pyodide {
     runPython: (code: string, globals?: any) => any;
@@ -64,19 +64,19 @@ class PythonWorker extends Backend {
     }
 
     override async _runCodeInternal(code: string): Promise<any> {
-        await this.pyodide.loadPackagesFromImports(code);
         if (this.initialized) {
             // run the code, potentially polluting the namespace
             // Functions and variables defined by the user become global
             // We need them to be global to let doctest work out of the box
-            try {
-                return await this.pyodide.globals.get(RUN_PYTHON_CODE)(code);
-            } finally {
-                // Cleanup the scope after computations are done
-                // Even in case of errors
-                this._cleanScope();
+            if (await this.pyodide.globals.get(PROCESS_PYTHON_CODE)(code, false)) {
+                await this.pyodide.loadPackagesFromImports(code);
+                await this.pyodide.globals.get(PROCESS_PYTHON_CODE)(code, true);
             }
+            // Cleanup the scope after computations are done
+            // Even in case of errors
+            this._cleanScope();
         } else {
+            await this.pyodide.loadPackagesFromImports(code);
             return this.pyodide.runPythonAsync(code);
         }
     }
--- a/src/workers/python/PythonWorker.worker.ts
+++ b/src/workers/python/PythonWorker.worker.ts
@@ -18,13 +18,11 @@ importScripts("https://cdn.jsdelivr.net/pyodide/v0.18.1/full/pyodide.js");
 class PythonWorker extends Backend {
     pyodide: Pyodide;
     initialized: boolean;
-    globals: Map<string, any>;
 
     constructor() {
         super();
         this.pyodide = {} as Pyodide;
         this.initialized = false;
-        this.globals = new Map();
     }
 
     override async launch(onEvent: (e: PapyrosEvent) => void,
@@ -41,40 +39,15 @@ class PythonWorker extends Backend {
             return this.onEvent(jsEvent);
         };
         this.pyodide.globals.get(INITIALIZE_PYTHON_BACKEND)(eventCallback);
-        this.globals = new Map((this.pyodide.globals as any).toJs());
         this.initialized = true;
-    }
-
-    _cleanScope(): void {
-        // Find the newly added globals
-        const pyodideGlobals = this.pyodide.globals;
-        const keysToRemove: Array<string> = [];
-        for (const key of pyodideGlobals.keys()) {
-            if (!this.globals.has(key)) {
-                keysToRemove.push(key);
-            } else {
-                // Reset value in case it was overriden
-                pyodideGlobals.set(key, this.globals.get(key));
-            }
-        }
-        // Remove them from the actual globals
-        // Separate runs of code should not be able to access variables/functions
-        // that were defined earlier on, as this could cause non-obvious bugs
-        keysToRemove.forEach(k => pyodideGlobals.delete(k));
     }
 
     override async _runCodeInternal(code: string): Promise<any> {
         if (this.initialized) {
-            // run the code, potentially polluting the namespace
-            // Functions and variables defined by the user become global
-            // We need them to be global to let doctest work out of the box
             if (await this.pyodide.globals.get(PROCESS_PYTHON_CODE)(code, false)) {
                 await this.pyodide.loadPackagesFromImports(code);
                 await this.pyodide.globals.get(PROCESS_PYTHON_CODE)(code, true);
             }
-            // Cleanup the scope after computations are done
-            // Even in case of errors
-            this._cleanScope();
         } else {
             await this.pyodide.loadPackagesFromImports(code);
             return this.pyodide.runPythonAsync(code);
--- a/src/workers/python/PythonWorker.worker.ts
+++ b/src/workers/python/PythonWorker.worker.ts
@@ -1,7 +1,7 @@
 import { expose } from "comlink";
 import { Backend } from "../../Backend";
 import { PapyrosEvent } from "../../PapyrosEvent";
-import { INITIALIZATION_CODE } from "./init.py";
+import { INITIALIZATION_CODE, INITIALIZE_PYTHON_BACKEND, RUN_PYTHON_CODE } from "./init.py";
 
 interface Pyodide {
     runPython: (code: string, globals?: any) => any;
@@ -40,7 +40,7 @@ class PythonWorker extends Backend {
             const jsEvent: PapyrosEvent = "toJs" in data ? data.toJs() : Object.fromEntries(data);
             return this.onEvent(jsEvent);
         };
-        this.pyodide.globals.get("__override_builtins")(eventCallback);
+        this.pyodide.globals.get(INITIALIZE_PYTHON_BACKEND)(eventCallback);
         this.globals = new Map((this.pyodide.globals as any).toJs());
         this.initialized = true;
     }
@@ -70,7 +70,7 @@ class PythonWorker extends Backend {
             // Functions and variables defined by the user become global
             // We need them to be global to let doctest work out of the box
             try {
-                return await this.pyodide.globals.get("__run_code")(code);
+                return await this.pyodide.globals.get(RUN_PYTHON_CODE)(code);
             } finally {
                 // Cleanup the scope after computations are done
                 // Even in case of errors
--- a/src/workers/python/PythonWorker.worker.ts
+++ b/src/workers/python/PythonWorker.worker.ts
@@ -1,6 +1,7 @@
 import { expose } from "comlink";
 import { Backend } from "../../Backend";
 import { PapyrosEvent } from "../../PapyrosEvent";
+import { LogType, papyrosLog } from "../../util/Logging";
 import { INITIALIZATION_CODE, INITIALIZE_PYTHON_BACKEND, PROCESS_PYTHON_CODE } from "./init.py";
 
 interface Pyodide {
@@ -43,13 +44,14 @@ class PythonWorker extends Backend {
     }
 
     override async _runCodeInternal(code: string): Promise<any> {
-        if (this.initialized) {
-            if (await this.pyodide.globals.get(PROCESS_PYTHON_CODE)(code, false)) {
-                await this.pyodide.loadPackagesFromImports(code);
-                await this.pyodide.globals.get(PROCESS_PYTHON_CODE)(code, true);
-            }
-        } else {
+        try {
             await this.pyodide.loadPackagesFromImports(code);
+        } catch (e) {
+            papyrosLog(LogType.Debug, "Something went wrong while loading imports: ", e);
+        }
+        if (this.initialized) {
+            await this.pyodide.globals.get(PROCESS_PYTHON_CODE)(code);
+        } else {
             return this.pyodide.runPythonAsync(code);
         }
     }
--- a/src/workers/python/PythonWorker.worker.ts
+++ b/src/workers/python/PythonWorker.worker.ts
@@ -71,7 +71,7 @@ class PythonWorker extends Backend {
             // Functions and variables defined by the user become global
             // We need them to be global to let doctest work out of the box
             try {
-                return this.pyodide.runPython(code);
+                return this.pyodide.globals.get("__run_code")(code);
             } finally {
                 // Cleanup the scope after computations are done
                 // Even in case of errors
--- a/src/workers/python/init.py.ts
+++ b/src/workers/python/init.py.ts
@@ -32,19 +32,28 @@ class __Papyros():
 
     def override_output(self):
         class __OutputWriter:
-            def __init__(self, type, on_write):
+            def __init__(self, type, on_write, original):
                 self.encoding = "utf-8"
                 self.type = type
                 self.on_write = on_write
+                self.original = original
                 
             def write(self, s):
+                if isinstance(s, bytes):
+                    s = s.decode("utf8", "replace")
                 self.on_write(dict(type=self.type, data=s))
 
             def flush(self):
                 pass # Given data is always immediately written
+
+            def __getattr__(self, item):
+                if item in self.__dict__:
+                    return self.item
+                return getattr(self.original, item)
+
         on_write = lambda d: self.message(d)
-        sys.stdout = __OutputWriter("output", on_write)
-        sys.stderr = __OutputWriter("error", on_write)
+        sys.stdout = __OutputWriter("output", on_write, sys.stdout)
+        sys.stderr = __OutputWriter("error", on_write, sys.stderr)
 
     def readline(self, n=-1, prompt=""):
         if not self.line:
--- a/src/workers/python/init.py.ts
+++ b/src/workers/python/init.py.ts
@@ -1,51 +1,72 @@
+export const INITIALIZE_PYTHON_BACKEND = "__init_papyros";
+export const RUN_PYTHON_CODE = "__run_code";
 export const INITIALIZATION_CODE =
     `
 from pyodide import to_js, eval_code_async
 import sys
+import micropip
+await micropip.install('friendly')
+import friendly
 
-def __override_builtins(cb):
-    __capture_stdout(cb)
-    __override_stdin(cb)
-    # set name to main instead of builtins
-    globals()["__name__"] = "__main__"
+__papyros = None
 
-def __capture_stdout(cb):
-    class _OutputWriter:
-        def __init__(self):
-            self.encoding = "utf-8"
-            
-        def write(self, s):
-            cb(to_js({"type": "output", "data":s}))
+class __Papyros():
+    def __init__(self, cb):
+        self.cb = cb
+        self.line = ""
+        self.override_builtins()
 
-        def flush(self):
-            pass # Given data is always immediately written
+    def message(self, data):
+        return self.cb(to_js(data))
 
-    sys.stdout = _OutputWriter()
+    def override_builtins(self):
+        self.override_output()
+        self.override_input()
 
-__papyros_input = ""
-def __override_stdin(cb):
+    def override_output(self):
+        class _OutputWriter:
+            def __init__(self, type, on_write):
+                self.encoding = "utf-8"
+                self.type = type
+                self.on_write = on_write
+                
+            def write(self, s):
+                self.on_write(dict(type=self.type, data=s))
 
-    def __papyros_input(prompt=""):
-        return __papyros_readline(prompt=prompt)[:-1] # Remove newline
+            def flush(self):
+                pass # Given data is always immediately written
 
-    def __papyros_readline(n=-1, prompt=""):
-        global __papyros_input
-        if not __papyros_input:
-            __papyros_input = cb(to_js({"type": "input", "data": prompt})) + "\\n"
-        if n < 0 or n > len(__papyros_input):
-            n = len(__papyros_input)
-        to_return = __papyros_input[0:n]
-        __papyros_input = __papyros_input[n:]
+        on_write = lambda d: self.message(d)
+        sys.stdout = _OutputWriter("output", on_write)
+        sys.stderr = _OutputWriter("error", on_write)
+
+    def readline(self, n=-1, prompt=""):
+        if not self.line:
+            self.line = self.message(dict(type="input", data=prompt)) + "\\n"
+        if n < 0 or n > len(self.line):
+            n = len(self.line)
+        to_return = self.line[0:n]
+        self.line = self.line[n:]
         return to_return
 
-    global input
-    input = __papyros_input
-    sys.stdin.readline = __papyros_readline
+    def override_input(self):
+        global input
+        input = lambda prompt="": self.readline(prompt=prompt)[:-1] # Remove newline
 
-async def __run_code(code, filename="code.py"):
+        sys.stdin.readline = self.readline
+
+def ${INITIALIZE_PYTHON_BACKEND}(cb):
+    globals()["__name__"] = "__main__"
+    global __papyros
+    __papyros = __Papyros(cb)
+
+async def ${RUN_PYTHON_CODE}(code, filename="my_code.py"):
     with open(filename, "w") as f:
         f.write(code)
     m = sys.modules.get("__main__")
     m.__file__ = filename
-    return await eval_code_async(code, globals(), filename=filename)
+    try:
+        return await eval_code_async(code, globals(), filename=filename)
+    except Exception:
+        friendly.explain_traceback()
 `;
--- a/src/workers/python/init.py.ts
+++ b/src/workers/python/init.py.ts
@@ -105,19 +105,14 @@ def ${INITIALIZE_PYTHON_BACKEND}(cb):
     global papyros
     papyros = Papyros(cb)
 
-async def ${PROCESS_PYTHON_CODE}(code, run, filename="my_code.py"):
+async def ${PROCESS_PYTHON_CODE}(code, filename="my_code.py"):
     with open(filename, "w") as f:
         f.write(code)
     friendly_traceback.source_cache.cache.add(filename, code)
     try:
-        if run:
-            await eval_code_async(code, papyros.globals(filename),
+        await eval_code_async(code, papyros.globals(filename),
                 filename=filename, return_mode="none")
-        else: # Only compile code (TODO separate Backend endpoint)
-            compile(code, filename, mode="exec", flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)
-        return True
     except Exception as e:
         papyros.message(dict(type="error", data=format_exception(filename, e)))
-        return False
 
 `;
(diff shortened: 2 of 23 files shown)


src/workers/python/init.py.ts
```diff
@@ -1,51 +1,72 @@
+export const INITIALIZE_PYTHON_BACKEND = "__init_papyros";
+export const RUN_PYTHON_CODE = "__run_code";
 export const INITIALIZATION_CODE =
     `
 from pyodide import to_js, eval_code_async
 import sys
+import micropip
+await micropip.install('friendly')
```
alexmojaki, line 8: Beware that this comes with several heavy dependencies. Overall it downloads some 2MB and takes a few seconds. Only starting this after pyodide has initialised is not ideal.

winniederidder: I switched to friendly_traceback, which should already cut some of these dependencies. According to https://pyodide.org/en/stable/usage/loading-packages.html there isn't really a way to load them even earlier, or am I mistaken here?

src/workers/python/init.py.ts
```diff
@@ -1,44 +1,118 @@
...
+    return html.escape(tb)
+
+def format_exception(filename, exc):
+    fr = FriendlyTraceback(type(exc), exc, exc.__traceback__)
+    fr.assign_generic()
+    fr.assign_cause()
+    tb = fr.info.get("simulated_python_traceback", "No traceback")
+    info = fr.info.get("generic", "No information available.")
+    why = fr.info.get("cause", "Unknown cause")
+    suggestions = fr.info.get("suggest", "No suggestions")
```
alexmojaki, line 75:
```suggestion
    suggestions = fr.info.get("suggest", "")
```

The dangling "No suggestions" looks weird to me. I imagine similar treatment for the other data might be good.

Screenshot from 2021-12-17 13-01-15

src/workers/python/init.py.ts
```diff
@@ -1,44 +1,118 @@
...
+    global __papyros
+    __papyros = __Papyros(cb)
+
+async def ${PROCESS_PYTHON_CODE}(code, run, filename="my_code.py"):
+    with open(filename, "w") as f:
+        f.write(code)
+    globals()["__file__"] = filename
+    try:
+        if run:
+            await eval_code_async(code, globals(), filename=filename)
```
alexmojaki, line 109:
```suggestion
            await eval_code_async(code, {}, filename=filename)
```

That should keep the scope clean.

src/workers/python/init.py.ts
```diff
@@ -1,44 +1,118 @@
...
+        global input
+        input = lambda prompt="": self.readline(prompt=prompt)[:-1] # Remove newline
+
+        sys.stdin.readline = self.readline
+
+def clean_traceback(tb, filename):
+    # For some reason the first character of __file__ is lost in friendly_traceback
+    for prefix in ["\\"", "\\'", " "]:
+        tb = tb.replace(prefix + filename[1:], prefix + filename)
+    return html.escape(tb)
```
alexmojaki, line 66: You actually need this for all your output, `print("<foo>")` doesn't work.

alexmojaki:
In Safari (14.1) running the same code with an exception twice gives a proper traceback the first time but a recursion error the second time.

src/workers/python/init.py.ts
alexmojaki: I explain what I do in futurecoder here: https://github.com/pyodide/pyodide/issues/1940#issuecomment-985041026

It's significantly more manual, but basically it starts downloading my code and dependencies in parallel with pyodide. Also I can write Python in a normal `.py` file, not in a giant JS string lol.

alexmojaki:
You can add translations for friendly-traceback [here](https://poeditor.com/join/project?hash=yC6iMTgLn5) which is nicer than manually editing a raw .po file.

I notice a few features missing from the traceback:

1. Highlighting the specific operation which failed, e.g. which part of `arr[i][j]` raised an error.
2. Showing all the lines of the expression/statement that failed (when it spans multiple lines)
3. Showing values of local variables in each frame.

Showing all this information is easy and won't require additional dependencies, although showing it in a visually appealing manner is another story. Which, if any, of these features do you want?

src/workers/python/init.py.ts
alexmojaki: Oh I see that messes with doctest, try something like this:

```python
import types
import sys
mod = types.ModuleType("__main__")
mod.__file__ = filename
sys.modules["__main__"] = mod
await eval_code_async(code, mod.__dict__, filename=filename)
```

pdawyndt:
I didn't know this "friendly" module, but definitely seems to be clarifying error messages for our target audience!

image

pdawyndt:
Some errors that we often see with students get a more natural explanation. I'll try to give some minimal examples where the default Python error messages are often way too cryptic for students:

### Adding two incompatible types

```
1 + []
```

### Extending a list with a non-iterable

```python
numbers = [1, 2, 3]
numbers += 4
```

### Add a mutable element to a set

```
numbers = [1, 2, 3]
{numbers}
```

But some are missed:

### Updating a set while iterating it

```python
numbers = {1, 2, 3}
for number in numbers:
    numbers.add(2 * number)
```

alexmojaki:
There are many Python errors out there! It's very actively developed and the author is very responsive, don't hesitate to make suggestions. I see one mention of that error: https://github.com/friendly-traceback/friendly-traceback/issues/45#issuecomment-962197860

I'm actually pleasantly surprised Python raises an exception at all, I wish it did that for lists, removing during iteration is a common mistake.

src/workers/python/init.py.ts
winniederidder: I added them temporarily as one of the top questions in this PR is what fallback to use. I'm waiting for my promotor to take a look and provide suggestions.

pdawyndt (approved):

winniederidder:
> You can add translations for friendly-traceback [here](https://poeditor.com/join/project?hash=yC6iMTgLn5) which is nicer than manually editing a raw .po file.

I am awaiting approval of both my promotors before starting on translating there.

> I notice a few features missing from the traceback:
> 
> 1. Highlighting the specific operation which failed, e.g. which part of `arr[i][j]` raised an error.
> 2. Showing all the lines of the expression/statement that failed (when it spans multiple lines)
> 3. Showing values of local variables in each frame.
> 
> Showing all this information is easy and won't require additional dependencies, although showing it in a visually appealing manner is another story. Which, if any, of these features do you want?

I think we are interested in all of these features.

> When I run this code:
> 
> ```python
> import micropip
> await micropip.install('friendly')
> import friendly
> try:
>     1/0
> except:
>     friendly.explain_traceback()
> ```
> 
> The console shows this error:
> 
> ```
> comlink.mjs:57 Uncaught (in promise) SyntaxError: Unexpected token T in JSON at position 1
>     at JSON.parse (<anonymous>)
>     at n.onError (bundle.js:1)
>     at n.onMessage (bundle.js:1)
>     at Function.<anonymous> (bundle.js:1)
>     at MessagePort.t (bundle.js:1)
> ```

I resolved this error by also accepting a string and just printing that. Now, I experienced on multiple occassions that friendly_traceback somehow deletes the first character of __file__ (e.g. it prints Exception raised on line 5 of file y_code.py.) which I find very confusing and I didn't find the reason by inspecting the source code of the library. Have you encountered this aswell?

alexmojaki:
> I am awaiting approval of both my promotors before starting on translating there.

I understand, just pointing it out for future reference.

> I think we are interested in all of these features.

Great! Just keep that in mind for now, no need for it to affect this PR. I'm planning on adding some more tooling to stack_data to make it easier to retrieve this info and use it in the frontend to construct your own interface.

> Now, I experienced on multiple occassions that friendly_traceback somehow deletes the first character of **file** (e.g. it prints Exception raised on line 5 of file y_code.py.) which I find very confusing and I didn't find the reason by inspecting the source code of the library. Have you encountered this aswell?

No I've never seen that, but I only use friendly_traceback for the error message explanations, I do the formatting myself. If you can isolate and reproduce the problem I'm sure a bug report would be well received.

Personally I'm most interested to see if you can get friendly_traceback working consistently on Safari without recursion errors.

winniederidder:
> Great! Just keep that in mind for now, no need for it to affect this PR. I'm planning on adding some more tooling to stack_data to make it easier to retrieve this info and use it in the frontend to construct your own interface.
> 
Sounds great!

> No I've never seen that, but I only use friendly_traceback for the error message explanations, I do the formatting myself. If you can isolate and reproduce the problem I'm sure a bug report would be well received.

To reproduce it, simply input your divide_by_zero example in Papyros. While \_\_file\_\_ is set correctly to my_code.py, the traceback shows it as y_code.py. I do not modify this string before displaying it.


> Personally I'm most interested to see if you can get friendly_traceback working consistently on Safari without recursion errors.
Is this an issue that friendly_traceback is aware of?  Didn't find any mention of Safari-bugs in their tracker. I also don't really have access to Safari, which makes solving issues related to that a bit more tricky.

alexmojaki:
> To reproduce it, simply input your divide_by_zero example in Papyros. While `__file__` is set correctly to `my_code.py`, the traceback shows it as `y_code.py`. I do not modify this string before displaying it.

I think I see the problem, I'll make an issue there.

> > Personally I'm most interested to see if you can get friendly_traceback working consistently on Safari without recursion errors.
> Is this an issue that friendly_traceback is aware of?  Didn't find any mention of Safari-bugs in their tracker. I also don't really have access to Safari, which makes solving issues related to that a bit more tricky.

This is probably just the difficulty of running Pyodide in a web worker in Safari. I was surprised to see that you got it working at all, [this comment](https://github.com/pyodide/pyodide/issues/441#issuecomment-678140870) made it sound like it was impossible to run *any* code, although I never tested it myself. See if there's ways to reduce the stack depth.

For the source caching issue, use these lines:

```python
        friendly_traceback.source_cache.cache.add(self.filename, source_code)
        stack_data.Source._class_local("__source_cache", {}).pop(self.filename, None)
```

The stack_data part should be automatic soon, see https://github.com/friendly-traceback/friendly-traceback/pull/96

src/workers/python/init.py.ts
```diff
@@ -11,6 +11,9 @@ import html
 import types
 import os
 import micropip
+# temporarily until friendly_traceback does this automatically
+await micropip.install("stack_data")
```
alexmojaki, line 15: `friendly_traceback` depends on `stack_data` (which is why you need to clear the stack_data cache) so you only need to install friendly_traceback, and import stack_data after that.

winniederidder:
> I think I see the problem, I'll make an issue there.

Thanks!

> This is probably just the difficulty of running Pyodide in a web worker in Safari. I was surprised to see that you got it working at all, [this comment](https://github.com/pyodide/pyodide/issues/441#issuecomment-678140870) made it sound like it was impossible to run _any_ code, although I never tested it myself. See if there's ways to reduce the stack depth.

In those discussions, you mention importing bigger modules first to limit the recursion depth. I guess this is something that can be played with, by e.g. importing friendly_traceback in smaller parts at a time. Not sure if this is something that can be done automatically and generically, so I'll take a look at that.

alexmojaki:
Safari seems to be able to import friendly traceback fine, the problems happen when using it. Have you tested it yourself? You can use browserstack to emulate Safari.

src/workers/python/init.py.ts
winniederidder: Since Safari doesn't handle those installs very well, this might be useful (I am considering giving other dependencies of friendly_traceback a similar treatment).

winniederidder:
> Safari seems to be able to import friendly traceback fine, the problems happen when using it. Have you tested it yourself? You can use browserstack to emulate Safari.

Browserstack does not seem to have a student plan, so I'm afraid that won't do. I can occasionally ask colleagues to give it a try, but those discussions mentioned Safari capping the stack at 50 while Pyodide expects a stack going up to 1000, so even if something will work, it'll likely be a less fleshed out version due to these constraints.

alexmojaki:
See if you can get a free open source plan: https://www.browserstack.com/open-source

I think I'm on that, although honestly I'm not sure.

winniederidder:
> Safari seems to be able to import friendly traceback fine, the problems happen when using it. Have you tested it yourself? You can use browserstack to emulate Safari.

By using Safari/Mac graciously provided by a peer, and the following simple testing code:
```python
def stackTest(i):
    print("Stack test number: " +str(i))
    stackTest(i+1)

stackTest(0)
```
I obtained a maximum stack size of 107 of Safari and 236 on Chrome.

winniederidder:
> See if you can get a free open source plan: https://www.browserstack.com/open-source
> 
> I think I'm on that, although honestly I'm not sure.

That indeed seems to work and will definitely help, thanks for this suggestion!

winniederidder:
> ```python
> import micropip
> await micropip.install('friendly')
> import friendly
> try:
>     1/0
> except:
>     friendly.explain_traceback()
> ```



> In Safari (14.1) running the same code with an exception twice gives a proper traceback the first time but a recursion error the second time.

I cannot seem to reproduce this issue on version 15. https://live.browserstack.com/dashboard#os=OS+X&os_version=Monterey&browser=Safari&browser_version=15.0&zoom_to_fit=true&full_screen=true&url=https%3A%2F%2Fdocs.dodona.be%2Fpapyros%2F&speed=1
Running code twice works, aswell as using friendly_traceback in Papyros itself.

alexmojaki:
> In Safari (14.1) running the same code with an exception twice gives a proper traceback the first time but a recursion error the second time.

Turns out the reproduction is more specific than I thought. I get the problem (including on version 15) when I run just the code `1/0` (nothing else) twice. Somehow it only seems to happen if that's the first thing I do, and not all exception types cause this.

aroberge:
The problem mentioned about friendly-traceback shortening the filename incorrectly by removing the first character should now be fixed  (friendly-traceback new version: 0.4.93).

winniederidder:
> Turns out the reproduction is more specific than I thought. I get the problem (including on version 15) when I run just the code `1/0` (nothing else) twice. Somehow it only seems to happen if that's the first thing I do, and not all exception types cause this.

Some more experimenting leads me to believe that as soon as the recursion depth has been exceeded one, no code will run anymore. It is also inconsistent as to when the recursion depth limit is reached. Doing a couple normal runs, then 1/0, then normal things again leads to no issues. A patch would be detecting this recursion error and restarting the backend or asking the user to refresh their page if these issues occur, but this is of course not ideal.

aroberge:
Currently, your site is using an obsolete version of friendly-traceback, possibly from my old repository (https://github.com/aroberge/friendly).  You should make sure to update **friendly-traceback** (and not **friendly) either directly from pypi.org (preferable) or from the new repository (https://github.com/friendly-traceback/friendly-traceback).

alexmojaki:
OK I found the problem by running the following code in papyros:

```python
import sys
import traceback

def trace(frame, event, arg):
    if event != 'call':
        return
    depth = 0
    while frame:
        depth += 1
        frame = frame.f_back

    if depth > 70:
        traceback.print_stack()
        raise ValueError

sys.settrace(trace)

1/0
```

The second time I run it (doesn't need to be in Safari) shows this stack when it gets to >70 frames:

```
File "/lib/python3.9/asyncio/events.py", line 80, in _run
self._context.run(self._callback, *self._args)
File "/lib/python3.9/asyncio/tasks.py", line 256, in __step
result = coro.send(None)
File "<exec>", line 111, in __process_code
File "/lib/python3.9/site-packages/_pyodide/_base.py", line 494, in eval_code_async
await CodeRunner(
File "/lib/python3.9/site-packages/_pyodide/_base.py", line 249, in compile
self._gen.send(self.ast)
File "/lib/python3.9/site-packages/_pyodide/_base.py", line 155, in _parse_and_compile_gen
_last_expr_to_raise(mod)
File "/lib/python3.9/site-packages/_pyodide/_base.py", line 116, in _last_expr_to_raise
raise_expr = deepcopy(_raise_template_ast)
File "/lib/python3.9/copy.py", line 172, in deepcopy
y = _reconstruct(x, memo, *rv)
File "/lib/python3.9/copy.py", line 270, in _reconstruct
state = deepcopy(state, memo)
File "/lib/python3.9/copy.py", line 146, in deepcopy
y = copier(x, memo)
File "/lib/python3.9/copy.py", line 230, in _deepcopy_dict
y[deepcopy(key, memo)] = deepcopy(value, memo)
File "/lib/python3.9/copy.py", line 172, in deepcopy
y = _reconstruct(x, memo, *rv)

[...skipping many deepcopy frames...]

File "/lib/python3.9/copy.py", line 172, in deepcopy
y = _reconstruct(x, memo, *rv)
File "/lib/python3.9/copy.py", line 264, in _reconstruct
y = func(*args)
File "/lib/python3.9/copy.py", line 263, in <genexpr>
args = (deepcopy(arg, memo) for arg in args)
File "/lib/python3.9/copy.py", line 146, in deepcopy
y = copier(x, memo)
File "/lib/python3.9/copy.py", line 210, in _deepcopy_tuple
y = [deepcopy(a, memo) for a in x]
File "/lib/python3.9/copy.py", line 210, in <listcomp>
y = [deepcopy(a, memo) for a in x]
File "my_code.py", line 13, in trace
traceback.print_stack()
```

So this is actually caused by pyodide's `eval_code_async`. Looking at the source, passing `return_mode="none"` should fix it, although I don't think you need `eval_code_async` at all.

@aroberge I don't know what you're looking at, the site is automatically using the latest version of friendly_traceback, which it installs on the fly. The filename issue is resolved, we're now talking about a different problem which I thought was caused by friendly_traceback but apparently isn't.

src/workers/python/init.py.ts
aroberge: "suggestions" should not be included.  when it exists, it is a shorter version of "why" and is only useful in interactive mode (at a Python console) to accompany the traceback.

"shortened_traceback" should be used instead of "simulated_python_traceback".

aroberge:
@alexmojaki On https://docs.dodona.be/papyros I generated a SyntaxError which was not recognized and got a message about filing an issue with the old url.  I just tried to recreate the SyntaxError and it is definitely recognized - so this tells me that a newer version is now used.

aroberge:
I added two suggestions https://github.com/dodona-edu/papyros/pull/80/files#r771985137 on the PR which, I think, should be included in future deployments.

src/workers/python/init.py.ts
alexmojaki: Well it can be removed now anyway

src/workers/python/init.py.ts
```diff
@@ -1,44 +1,127 @@
...
-    input = __papyros_input
-    sys.stdin.readline = __papyros_readline
+    def override_input(self):
+        global input
+        input = lambda prompt="": self.readline(prompt=prompt)[:-1] # Remove newline
+
+        sys.stdin.readline = self.readline
+
+def clean_traceback(tb, filename):
+    # For some reason the first character of __file__ is lost in friendly_traceback
```
alexmojaki, line 68: Should be able to remove this now

src/workers/python/PythonWorker.worker.ts
```diff
@@ -65,16 +64,19 @@ class PythonWorker extends Backend {
     }
 
     override async _runCodeInternal(code: string): Promise<any> {
-        await this.pyodide.loadPackagesFromImports(code);
         if (this.initialized) {
             // run the code, potentially polluting the namespace
             // Functions and variables defined by the user become global
             // We need them to be global to let doctest work out of the box
```
alexmojaki, line 70: These can be removed :tada:

src/workers/python/PythonWorker.worker.ts
```diff
@@ -65,16 +64,19 @@ class PythonWorker extends Backend {
...
-            const result = this.pyodide.runPython(code);
+            if (await this.pyodide.globals.get(PROCESS_PYTHON_CODE)(code, false)) {
+                await this.pyodide.loadPackagesFromImports(code);
```
alexmojaki, line 72: I had a similar issue with `find_imports` on invalid syntax code, which led to a fix from pyodide: https://github.com/pyodide/pyodide/pull/1819

However it doesn't seem to actually be released which is weird: https://github.com/pyodide/pyodide/pull/1820

Anyway, I think it'd be better to instead wrap `loadPackagesFromImports` in a `try/catch` because if anything goes wrong it's probably not worth breaking the whole application anyway for the sake of automatic import installing. That's what I do at least.

aroberge:
The following is a draft explanation for the currently unhandled example mentioned in https://github.com/dodona-edu/papyros/pull/80#issuecomment-996880621

```
Traceback (most recent call last):
  File "HOME:\github\friendly-traceback\ignore.py", line 2, in <module>
    for number in numbers:
RuntimeError: Set changed size during iteration

    A `RuntimeError` is raised when an error is detected that doesn't fall in any
    of the more specific exception types defined by Python.

    While you were iterating over the elements of `numbers` (a `set`)
    in a `for` or `while` loop, you either tried to add or remove items from it.
    Suggestion: start by making a copy of `numbers` and iterate over the elements
    of that copy if you want to change `numbers` inside a loop.

    Exception raised on line 2 of file HOME:\github\friendly-traceback\ignore.py.

       1: numbers = {1, 2, 3}
    -->2: for number in numbers:
       3:     numbers.add(2 * number)

            number:  2
            numbers:  {1, 2, 3, 4}
```
If this meets your approval, it can be added to the next release of friendly-traceback.  Feel free to make suggestions for better wording.

src/workers/python/PythonWorker.worker.ts
winniederidder: Required a small change to ensure input was also correctly overriden, but this has been done now 🎉

src/workers/python/PythonWorker.worker.ts
winniederidder: I do want friendly_traceback to intercept the potential syntax errors, which is why I run it through compile first. The pyodide package-loading code builds the AST to find import statements, so in order to catch these compile errors I need to do this. Unless you mean moving this part into Python aswell, which might not be a bad idea? :p

winniederidder:
> The following is a draft explanation for the currently unhandled example mentioned in [#80 (comment)](https://github.com/dodona-edu/papyros/pull/80#issuecomment-996880621)
> 
> ```
> Traceback (most recent call last):
>   File "HOME:\github\friendly-traceback\ignore.py", line 2, in <module>
>     for number in numbers:
> RuntimeError: Set changed size during iteration
> 
>     A `RuntimeError` is raised when an error is detected that doesn't fall in any
>     of the more specific exception types defined by Python.
> 
>     While you were iterating over the elements of `numbers` (a `set`)
>     in a `for` or `while` loop, you either tried to add or remove items from it.
>     Suggestion: start by making a copy of `numbers` and iterate over the elements
>     of that copy if you want to change `numbers` inside a loop.
> 
>     Exception raised on line 2 of file HOME:\github\friendly-traceback\ignore.py.
> 
>        1: numbers = {1, 2, 3}
>     -->2: for number in numbers:
>        3:     numbers.add(2 * number)
> 
>             number:  2
>             numbers:  {1, 2, 3, 4}
> ```
> 
> If this meets your approval, it can be added to the next release of friendly-traceback. Feel free to make suggestions for better wording.

@pdawyndt

src/workers/python/init.py.ts
```diff
@@ -1,44 +1,116 @@
...
         return to_return
 
-    global input
-    input = __papyros_input
-    sys.stdin.readline = __papyros_readline
+    def globals(self, filename):
+        mod = types.ModuleType("__main__")
+        mod.__file__ = filename
+        sys.modules["__main__"] = mod
+        mod.input = lambda prompt="": self.readline(prompt=prompt)[:-1] # Remove newline
```
alexmojaki, line 62: This needs to be set on `builtins` (that's where it is normally) to work from everywhere, in case some other function not in the user's code calls `input()`.

src/workers/python/init.py.ts
```diff
@@ -1,44 +1,116 @@
...
+        return self.cb(to_js(data))
+
+    def override_builtins(self):
+        self.override_output()
+        self.override_input()
+
+    def override_output(self):
+        class __OutputWriter:
+            def __init__(self, type, on_write):
+                self.encoding = "utf-8"
```
alexmojaki, line 36: This code:

```
import pytest

pytest.main()
```

gives a traceback ending in:

```
INTERNALERROR> File "/lib/python3.9/site-packages/_pytest/terminal.py", line 228, in pytest_configure
INTERNALERROR> reporter = TerminalReporter(config, sys.stdout)
INTERNALERROR> File "/lib/python3.9/site-packages/_pytest/terminal.py", line 337, in __init__
INTERNALERROR> self.isatty = file.isatty()
INTERNALERROR> AttributeError: '__OutputWriter' object has no attribute 'isatty'
```

To deal with `isatty`, `encoding`, `flush`, and any other weird attributes code might access, I use this code:

```
    def __getattr__(self, item):
        return getattr(sys.__stdout__, item)
```

src/workers/python/init.py.ts
```diff
@@ -1,44 +1,116 @@
...
+                self.type = type
+                self.on_write = on_write
+                
+            def write(self, s):
+                self.on_write(dict(type=self.type, data=s))
```
alexmojaki, line 41:
```suggestion
                if isinstance(s, bytes):
                    s = s.decode("utf8", "replace")
                self.on_write(dict(type=self.type, data=s))
```

Some libraries (e.g. flask) write bytes directly to stdout/stderr.

winniederidder:
@alexmojaki Thanks a lot for your detailed feedback and insights, they certainly help to improve this project!

src/workers/python/init.py.ts
```diff
@@ -32,19 +32,28 @@ class __Papyros():
...
+                if isinstance(s, bytes):
+                    s = s.decode("utf8", "replace")
                 self.on_write(dict(type=self.type, data=s))
 
             def flush(self):
                 pass # Given data is always immediately written
+
+            def __getattr__(self, item):
+                if item in self.__dict__:
+                    return self.item
```
alexmojaki, line 51: `__getattr__` is only for attributes which couldn't be resolved normally, so this would never happen and you don't need this. But if it did, `self.item` wouldn't work.

pdawyndt:
> The following is a draft explanation for the currently unhandled example mentioned in [#80 (comment)](https://github.com/dodona-edu/papyros/pull/80#issuecomment-996880621)
> 
> ```
> Traceback (most recent call last):
>   File "HOME:\github\friendly-traceback\ignore.py", line 2, in <module>
>     for number in numbers:
> RuntimeError: Set changed size during iteration
> 
>     A `RuntimeError` is raised when an error is detected that doesn't fall in any
>     of the more specific exception types defined by Python.
> 
>     While you were iterating over the elements of `numbers` (a `set`)
>     in a `for` or `while` loop, you either tried to add or remove items from it.
>     Suggestion: start by making a copy of `numbers` and iterate over the elements
>     of that copy if you want to change `numbers` inside a loop.
> 
>     Exception raised on line 2 of file HOME:\github\friendly-traceback\ignore.py.
> 
>        1: numbers = {1, 2, 3}
>     -->2: for number in numbers:
>        3:     numbers.add(2 * number)
> 
>             number:  2
>             numbers:  {1, 2, 3, 4}
> ```
> 
> If this meets your approval, it can be added to the next release of friendly-traceback. Feel free to make suggestions for better wording.

@aroberge This will definitely clarify the issue to students!

src/workers/python/PythonWorker.worker.ts
alexmojaki: Doesn't `eval_code_async` raise the SyntaxError so you can still handle it there?

src/workers/python/init.py.ts
```diff
@@ -1,44 +1,124 @@
...
+            what=what
+        )
+    )
+
+def ${INITIALIZE_PYTHON_BACKEND}(cb):
+    global __papyros
+    __papyros = __Papyros(cb)
+
+async def ${PROCESS_PYTHON_CODE}(code, run, filename="my_code.py"):
+    global __papyros
```
alexmojaki, line 109: `global` is not needed when you're only reading the variable.

Also these names don't need to start with `__` any more.

src/workers/python/PythonWorker.worker.ts
winniederidder: eval_code_async would also raise an Error, but I don't want it to try running the code as the imports wouldn't be available yet, so the raised Error could be the wrong one. Just compiling also sounds more efficient and clear.

alexmojaki: I don't think you understand. I'm suggesting this:

```js
try {
  await this.pyodide.loadPackagesFromImports(code);
} catch (e) {
  // Any failure to load packages shouldn't break everything
  console.error(e);
}
await this.pyodide.globals.get(PROCESS_PYTHON_CODE)(code);
```

Then you can also remove the boolean `run` argument and the associated conditional.

winniederidder: This does not allow me to have friendly_traceback take care of the syntax error that could occur during importing. I know loadPackagesFromImports does a call to compile, but I want students to be able to understand where it went wrong. I believe loadPackages does not load any modules if a Syntax error occurred. However, in actual code, I think it is possible for the Syntax error to occur after a module was loaded and used. Would eval_code_async not give a MdouleNotFoundError instead of the Syntax error? This is why I want the compilation step to happen beforehand.

alexmojaki: If `code` contains a syntax error, `loadPackagesFromImports` will fail to load anything, and `eval_code_async` will raise the syntax error for friendly_traceback to catch.

If `code` doesn't contain a syntax error, `loadPackagesFromImports` might load something, and `eval_code_async` will not raise the syntax error for `code` itself. It might raise a syntax error when doing the actual import if the imported code contains invalid syntax, but you will still catch that as usual. It won't raise ModuleNotFoundError but I don't think that matters.

Running `loadPackagesFromImports` doesn't interfere with anything. The code will be compiled/parsed twice regardless of whether there's an error.

alexmojaki:
Matplotlib is looking pretty cool! As is tradition by now, it doesn't work in Safari. Looks like a known issue in pyodide: https://github.com/pyodide/pyodide/issues/1921

winniederidder:
> Matplotlib is looking pretty cool! As is tradition by now, it doesn't work in Safari. Looks like a known issue in pyodide: [pyodide/pyodide#1921](https://github.com/pyodide/pyodide/issues/1921)

Indeed, that is quite the bummer. I used the workaround from https://github.com/pyodide/pyodide/issues/1518 but it didn't mention anything Safari related. I'll wrap it in a try-catch then.

src/workers/python/init.py.ts
```diff
@@ -1,44 +1,140 @@
...
 import sys
+import ast
+import json
+import html
+import types
+import os
+import micropip
+await micropip.install('friendly_traceback')
+import friendly_traceback
+from friendly_traceback.core import FriendlyTraceback
```
alexmojaki, line 16: Both installing and simply importing friendly_traceback affect startup significantly. Try starting the installation here but only awaiting the installation, importing, and updating the source cache when an exception is raised.

bmesuere (approved):
I propose you merge this PR since it's probably beyond fixing ;) As I said before, after the proof of concept integration, we'll have to take a look at how the code is structured (e.g. the python code strings mentioned by @alexmojaki, list what works and doesn't in Safari, document the hacks we use and decide if we want to support it, etc) 

I would suggest you update the PR start with the additional changes made here (such as the examples).

Regarding your questions:
- The UI can be improved but I suggest we keep that for a later time
- If fields are missing handle that in the ui
- Dutch translations for the errors would be nice but is not a priority. Getting this proof of concept integrated in Dodona is.

PR #92, opened 2022-01-17
Use webpack raw-loader to keep Python code in an actual .py file

alexmojaki:
Personally I find it much nicer and easier to work with a `.py` file that my IDE can understand properly rather than a big string literal in a `.ts` file. I hope you'll agree.


winniederidder:
It had the benefit of being able to share JavaScript constants (such as the code entry-points like init_papyros). However, using it for Python also allows using it for other languages that also might have setup code that can benefit from IDE support.

winniederidder (approved):

Issue #63, opened 2021-12-03
futurecoder


alexmojaki:
[futurecoder](https://futurecoder.io/) has solved many of the problems you're working on. It has an interactive REPL, beginner-friendly tracebacks and linting, and integrated debuggers. The framework for exercises also has many useful features, although I haven't compared to dodona.

What do you think of merging futurecoder and papyros, or integrating the futurecoder IDE into dodona instead of papyros? Then we could combine our efforts to add whatever features you find to be missing from futurecoder, which would hopefully be less work than what you have to do now and would benefit you in the long term.

bmesuere:
Hi Alex,

Thank you for your interest in our project. As you have already figured out, our main project is [Dodona](https://dodona.ugent.be/) (you should be able to sign in if you have an office 365 of gsuite domain). Our focus there is an exercise platform, specifically targeted on secondary and higher education. Teachers can create a course and select from our 8000 exercises and easily follow up the progress of their students. Code submissions of students are queued in our execution environment and run through tests defined by the creator of the exercise. The students then get almost instant feedback on their submissions. Dodona is used extensively at Ghent University, but is also seeing usage from almost all institutions for higher education in Flanders. Since this year, we're seeing an uptake in usage from secondary schools as well. The reason for this is that "computational thinking" is now a mandatory topic in the second grade and many school book publishers use Dodona for their exercises. We not only support Python, but also Java, JavaScript, Haskell, Prolog, C, C#, R, Kotlin, Scheme, mysql and html.

While Dodona works great, many secondary schools have let us know that it is not always straightforward for them to run Python code locally since it-admins are reluctant to install Python and an IDE on school computers. (In my opinion, a full IDE is also not a great match for 14-16 year olds trying Python for the first time). This is why we decided to develop a light-weight alternative to run (basic) Python code in the browser. The goal was then to both offer this as a stand-alone tool/web page, as well as integrate it in our Dodona platform as a "scratchpad" to quickly run some code without submitting a solution. @winniederidder, who has done most of the coding so far, is developing this as his graduation project for his master's degree.

[...]

Issue #17, opened 2021-11-14
Issues with lots of output


pdawyndt:
The code

```python
for number in range(10000):
    print(number)
```

works the first time you run it on Papyros. However, Papyros stalls the second time you run it. You can still "Terminate" the code though, and run something else. That works fine.

winniederidder:
I've done some experimenting regarding this issue on https://github.com/dodona-edu/papyros/tree/issue17/too-much-output. 
First off, I noticed indeed that the first run works. Looking at the logs, when executing a second time, the code is sent to the worker but not received, so it is not executed. However, after a long time (around 211 seconds), the message did get through. I'm not sure why it blocks, as all the output messages have been successfully delivered.

When only printing 1000 numbers instead of 10000, the problem does not arise, so I figured it has something to do with the amount of messages sent back and forth. In the mentioned branch, I added a queueing system so that print messages are buffered before being sent over to the main thread. However, queue sizes of 100 or 1000 did not solve the problem, the second run still did not get through in a timely fashion.

A potential alternative is just ignoring too much output and notifying the user that further output is not displayed. Perhaps even writing to a file, something Jupyter notebooks do when too much output is generated. But at this moment I don't see a proper solution as I found no helpful information on the internet either regarding too many messages or data being sent.

winniederidder:
@bmesuere @pdawyndt Thoughts on my writings here?

pdawyndt:
We cannot simply ignore lots of output not working. Given the target audience, this will occur.

winniederidder:
I agree, but my research showed that the run-code message seemed to be delayed. I am unsure why this is, and the queueing system I implemented did not solve the issue. The issue is also present in the JavaScript version, so it is not related to Pyodide. It is thus likely to be a Comlink issue, or even something regarding the Messaging system used by WebWorkers (a data or message limit? But queueing would solve the message limit, and I did not find any other clues).
I think we would like to keep displaying messages while running the code, so I can propose the following solutions:
- Hacky and simple: if the code hasn't started running a couple of seconds in, restart the backend and retry.
- Try a file-based solution where large outputs are written to a file instead of displayed (like Jupyter notebooks)
- Ignore displaying long outputs and simply show a truncation warning.

alexmojaki:
I 'solved' this for Python, was it solved elsewhere for JS?

winniederidder:
Indeed, JavaScript still suffers from this issue. Could you elaborate a bit more on how you tackled this problem internally?

[...]

PR #96, opened 2022-02-04
Use python_runner

alexmojaki:
Part 1 of #63. Uses https://github.com/alexmojaki/python_runner which takes care of things like input and output and is used by futurecoder.

`python_runner` buffers output internally which helps a lot with #17.

Also in this PR:

- Replace `\\n` (introduced in my previous PR #92) with `\n`
- Only install `matplotlib` when it's imported by the user's code, greatly speeding up startup.


winniederidder (approved):
Decreases startup time by a lot, which is wonderful. The first matplotlib run is now longer of course, but this isn't our main use case and we definitely don't want it to impact everyone all the time.

PR #102, opened 2022-02-21
Show modules being loaded

### LINKED ISSUE #98: Show modules being loaded in statusbar
Now that we swithed to lazy loading of non-standard modules like `matplotlib`, it takes some extra time to run code that initializes the process of loading the module. Instead of using the message "Running" in the statusbar throughout the entire process, we might have a spefic message while loading modules (maybe even specifying which module is being loaded) and not take that time into account when measuring the runtime.

winniederidder:
Depends on #101 

Adds an idicator to which package is currently being installed.
papyros-installing

Downside: some modules require many packages (of which some are quite small) which might cause flashing in text. Especially notable in the matplotlib example @pdawyndt 

Fixes #98

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -25,3 +25,5 @@ export const DEFAULT_LOCALE = "nl";
 
 export const INPUT_RELATIVE_URL = "__papyros_input";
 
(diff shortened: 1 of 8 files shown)


winniederidder:
Has been resolved in a better way in #178

PR #104, opened 2022-02-26
Use sync-message for input

### LINKED ISSUE #95: Service worker stops working in Safari after refresh
When I test Safari in Browserstack, I find that `input()` works fine at first, but stops working after I reload the page. I encountered the same problem while working on a generic synchronous communication library. It's not entirely clear but it seems like Safari service workers stop intercepting synchronous (or maybe all?) XHR requests after the reload. Can anyone with an actual Mac confirm this issue? Anyone know anything more?

alexmojaki:
Part of #63 

Fixes #95 and #27 (the latter shouldn't have been closed)

Also fixes another bug not in any issue: if two open tabs are both reading input from the service worker, then they will interfere with each other. This is because a single service worker is always shared between all open tabs, and the previous service worker only had one place to store input.

### DIFF
--- a/src/InputManager.ts
+++ b/src/InputManager.ts
@@ -2,12 +2,12 @@ import { t } from "i18n-js";
 import {
     SWITCH_INPUT_MODE_A_ID,
     INPUT_AREA_WRAPPER_ID,
-    INPUT_RELATIVE_URL, INPUT_TA_ID, SEND_INPUT_BTN_ID
+    INPUT_TA_ID, SEND_INPUT_BTN_ID
 } from "./Constants";
 import { PapyrosEvent } from "./PapyrosEvent";
 import { papyrosLog, LogType } from "./util/Logging";
 import { addListener, RenderOptions, renderWithOptions } from "./util/Util";
-
+import { Channel, writeMessage } from "sync-message";
 
 export enum InputMode {
     Interactive = "interactive",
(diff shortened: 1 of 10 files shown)


winniederidder:
Any plans on providing documentation on both python_runner and sync_message to get a better overview of what they can accomplish and how to use them properly?  The library contains more functionality than I had provided myself so I'd like to fully understand what's going on.

alexmojaki:
There will definitely be documentation, I just wanted to test it out in practice so that the API is nice and stable first. Do you want that now? I don't think there's anything special you need to know at this stage. I'm already working on the next layers which will deal with some details.

alexmojaki:
Let me know when you no longer plan on making changes to this area of the code and you're close to being able to merge this PR, and I'll resolve the conflicts then.

src/InputManager.ts
```diff
@@ -28,30 +28,14 @@ export class InputManager {
...
-            // 2 Int32s:
-            // index 0 indicates whether data is written
-            // index 1 denotes length of the string
-            this.inputMetaData = new Int32Array(
-                new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT * 2)
-            );
-        } else {
-            papyrosLog(LogType.Important, "Using serviceWorker for input");
-        }
+        this.channel = makeChannel({serviceWorker: {scope: window.location.href + "/"}})!;
```
alexmojaki, line 38: Replace `window.location.href` with `serviceWorkerRoot` or `inputURL` or whatever from your most recent commits. `scope` should be a URL that is *within* the scope of the registered service worker, so that sync-message can make requests to `scope` plus an extra path and know that they will be intercepted.

winniederidder: I'm merging this myself to ensure it all works properly within Dodona itself. For ease of use, the location of the service worker should be configurable while still allowing the worker to do all its tasks.

winniederidder:
I'm merging this myself to ensure it all works properly within Dodona itself. For ease of use, the location of the service worker should be configurable while still allowing the worker to do all its tasks.

alexmojaki:
> I'm merging this myself

As in, you're going to make the code changes again manually?

> For ease of use, the location of the service worker should be configurable while still allowing the worker to do all its tasks.

That's fine, the comment about `scope` handles that.

alexmojaki:
Added docs to https://github.com/alexmojaki/sync-message

PR #107, opened 2022-03-03
Documentation and refactoring

### LINKED ISSUE #58: Improve styling of output area
Currently, all data in the output area is treated equally. It would be helpful to use different colors for input, output, errors and so forth when displaying those kinds of messages.

### LINKED ISSUE #70: Disable input upon terminate
If a running program is terminated while it is waiting on interactive input, it should stop waiting for interactive input (clear and disable interactive input field).

Example code:

```python
for _ in range(3):
    name = input('Enter a name: ')
    print(f'Hello, {name}!')
```

Run this code, fill up the interactive input field with some text (without enter) and terminate. Text remains in place and can still be edited.

### LINKED ISSUE #75: Allow empty input
Python normally allows reading an empty line of stdin, in which case `input()` returns `''`. Neither interactive nor batch mode input seem to allow this in papyros. In particular I find the behaviour a bit surprising when the batch input starts with a blank line followed by a nonempty line.

### LINKED ISSUE #100: Simplify input handling
The service worker makes various things quite complicated. Support for SharedArrayBuffers is also growing. It would be better to fully switch over to SharedArrayBuffers to make everything work simply and efficiently.
Adding the required headers at the server side to support this will be needed. 
More specifically, the Cross-Origin-Embedder-Policy header must be set to 'require-corp', and the Cross-Origin-Resource-Policy must be set correctly (preferably the least broad option) to still allow fetching resources to work correctly. For example in Dodona, the sandboxed iframes must still be able to load code and stylesheets even when isolated.

winniederidder:
Some parts of the code were getting rather old and could be improved upon drastically thanks to new knowledge.
Other parts required documentation or a different approach (such as the InputManager).
To keep things manageable, PapyrosEvents now also support a contentType property to aid in parsing, as it was not always clear whether plain text, JSON or even images were being passed.

Closes #58 
Closes #70 
Closes #75 
Closes #100


bmesuere (approved):
This is a great improvement 👍

PR #108, opened 2022-03-08
Code completion

### LINKED ISSUE #40: Support code completion
Support code completion in Python code editor.

winniederidder:
Adds better autocompletion suggestions to the user while writing code.
Tasks:
- [x] Python support using Jedi
- [x] Correct sort order and show docstring if available

Fixes #40 
Depends on #107


src/workers/python/init.py
```diff
@@ -42,7 +42,7 @@ def show():
     async def run_async(self, source_code, mode="exec", top_level_await=True):
         """
         Mostly a copy of the parent `run_async` with `await ft` in case of an exception,
-        because `serialize_traceback` isn't async.
+        because `serialize_traceback` isn"t async.
```
pdawyndt, line 45: "isn't" should stay as is

winniederidder: The dangers of replace-all, will fix.

winniederidder:
Example of autocompletion on a module
autocomplete

Main issue is that it is quite slow for larger libraries (for example numpy). It's only slow the first time, after that caching occurs.
I also added auto-importing while typing to enable autocompletion before running any code.
@alexmojaki Do you have any plans/interests regarding autocompletion? The jedi module is fully functional within Pyodide but sadly a bit slow as mentioned, so I would welcome any suggestions.

alexmojaki:
No, I don't plan on adding autocomplete to futurecoder, and I have no experience with it, so I can't offer any tips about jedi.

pdawyndt:
Is there a short key to get "info" (e.g. on a built-in function) that has already been included in the code?

winniederidder:
> Is there a short key to get "info" (e.g. on a built-in function) that has already been included in the code?

At this moment, no. It does sound like a useful addition so I will take a look at it.

winniederidder:
> Is there a short key to get "info" (e.g. on a built-in function) that has already been included in the code?

Update: the default "initiate autocomplete"-keybind is Ctrl-Space, which can be used to start autocompleting if nothing is shown at the moment. This can e.g. autocomplete on the empty string to allow going through globals, but can also be used on already completed functions to show their info again. So this is sufficient to implement the suggestion. The keybind can be configured, but we might want to look into a general help section to list supported keybinds.

pdawyndt:
Does CodeMirror support resizable windows to display help. Currently the window displaying the online help cannot be resized, which requires a lot of scrolling.

winniederidder:
@pdawyndt I enabled resizing the box.

src/Backend.ts
```diff
@@ -87,8 +107,10 @@ export abstract class Backend {
     /**
      * Converts the context to a cloneable object containing useful properties
      * to generate autocompletion suggestions with
+     * Class instances are not passable to workers, so we extract the useful infromation
```
pdawyndt, line 110: Information

src/Backend.ts
```diff
@@ -68,4 +103,36 @@ export abstract class Backend {
         this.runId = runId;
         return await this._runCodeInternal(code);
     }
+
+    /**
+     * Converts the context to a cloneable object containing useful properties
+     * to generate autocompletion suggestions with
+     * Class instances are not passable to workers, so we extract the useful infromation
```
bmesuere, line 110:
```suggestion
     * Class instances are not passable to workers, so we extract the useful information
```

bmesuere (approved):

PR #111, opened 2022-03-27
Support linting

winniederidder:
This PR adds support for linting the user's code.
Choosing a library for linting was not trivial. They are usually meant to be used as a command-line tool which makes running them on the client side less practical. They also tend to use multiprocessing functionalities which is not fully supported by Pyodide. 

PyLint is a popular library but it cannot be imported in Pyodide due to not being a full Python package.
Flake8 can be imported, but needed some patching around its import calls to multiprocessing to even be able to run it.

UI-wise, tooltips in the gutter for lines with violations seem to be most appropriate and feasible.

### DIFF
--- a/src/workers/python/init.py
+++ b/src/workers/python/init.py
@@ -1,6 +1,7 @@
 import sys
 import json
 import os
+import builtins
 from collections.abc import Awaitable
 
 import micropip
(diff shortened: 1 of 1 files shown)


alexmojaki:
You can have arbitrary dependencies, but you have to package them beforehand rather than using micropip. pyodide-worker-runner helps with installing them.

winniederidder:
Closed and deleted in favor of #114

PR #112, opened 2022-03-30
Upgrade workers

### LINKED ISSUE #38: Avoid reloading the backend upon terminate
Upon clicking the `Terminate` button, the backend is reloaded. This does not seem necessary.

### LINKED ISSUE #66: No support for sleep
The `time.sleep()

### LINKED ISSUE #86: Enable Run button before loading completes
Often I have code ready to run before the backend has finished loading, e.g. when I have some code on the clipboard, or I want to run an example, or I stopped a running program. It's annoying having to wait for the Run button to become enabled. I think it should be clickable immediately, but the "Loading" spinner would still show and execution would only start when it completes, at which point the spinner would immediately transition to "Running". This likely only requires awaiting a particular promise.

winniederidder:
This PR upgrades the workers to make use of the recently published packages Comsync and pyodide-worker-runner.
It also adds useful refactoring parts from #109 that are not strictly tied into debugging but made integrating it more feasible.
This should increase the general flexibility.

Support was added for the following:
sleeping
interrupting with restarting
clicking run before the load is finished (is awaited during run instead if needed)
generating a tar with dependencies to pre-load, allowing internal Python code to use proper packages instead of a raw string
Running code is refactored to be done within a separate component contained within the Papyros object

Closes #66 
Closes #38 
Closes #86


src/CodeRunner.ts
```diff
@@ -114,34 +114,23 @@ export class CodeRunner {
...
-        this.publishEvent({
+        BackendManager.publish({
             type: BackendEventType.End,
-            runId: this.runId,
             data: "User cancelled run", contentType: "text/plain"
         });
         // Since we use workers, the old one must be entirely replaced to interrupt it
-        BackendManager.stopBackend(this.backend);
+        this.backend.interrupt();
         return this.start();
```
alexmojaki, line 124: Comsync handles restarting the worker. Also note that it's no longer necessarily replacing the worker.

src/BackendEvent.ts
```diff
@@ -24,11 +24,6 @@ export interface BackendEvent {
      * The type of action generating this event
      */
     type: BackendEventType;
-    /**
-     * The identifier for the run this message is associated with
-     * This allows discarding outdated events that were delayed
-     */
-    runId: number;
```
alexmojaki: Comsync only handles this for its own internal callbacks, you may still want this for your own events.

src/CodeRunner.ts
winniederidder: Replacing is no longer done. But there's probably more parts of start that should no longer be required, so I'll definitely look further into that.

src/BackendEvent.ts
winniederidder: The runId was only used to filter out delayed events. The interrupts should smoothen this process so that I no longer need this filtering step.

src/workers/python/PythonWorker.worker.ts
```diff
@@ -20,58 +21,48 @@ declare function loadPyodide(args: { indexURL: string; fullStdLib: boolean }): P
...
         let converted = data;
-        if ("toJs" in data) {
-            converted = data.toJs();
+        if (data.toJs) {
+            converted = data.toJs({ dict_converter: Object.fromEntries });
         }
-        return Object.fromEntries(converted);
+        if (converted instanceof Map) {
+            converted = Object.fromEntries(converted);
+        }
```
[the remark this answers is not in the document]
winniederidder: In theory yes, but I was still getting Map-objects at some point. I'll take a second look.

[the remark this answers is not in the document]
winniederidder: After some testing, this is still required. It seems like that if the top-object is a dictionary, you still end up with a Map instead of an object. I don't know if this is something we want to raise with Pyodide or if this is acceptable.

src/util/Util.ts
```diff
@@ -205,7 +205,13 @@ export function parseData(data: unknown, contentType: string): any {
...
                 case "json": {
-                    return JSON.parse(data as string);
+                    return JSON.parse(data);
+                }
+                case "integer": {
+                    return parseInt(data);
+                }
+                case "float": {
+                    return parseFloat(data);
                 }
```
alexmojaki, line 215: I don't understand why you put data in strings and then parse it, except for maybe the base64. Can't you transfer objects, arrays, and numbers as they are?

src/workers/python/init.py
```diff
@@ -55,6 +56,8 @@ def cb(typ, dat, **kwargs):
                         raise ValueError(f"Unknown output part type {typ}")
             elif event_type == "input":
                 return cb("input", data["prompt"], contentType="text/plain")
+            elif event_type == "sleep":
+                return cb("sleep", str(data["seconds"]*1000), contentType="text/integer")
```
alexmojaki, line 60: I suggest using `makeRunnerCallback` from pyodide-worker-runner and moving this event translation logic to JS.

src/workers/python/init.py
```diff
@@ -9,19 +11,79 @@
...
+            else:
+                raise ValueError(f"Unknown event type {event_type}")
+
+        self.set_callback(runner_callback)
+
+    def set_file(self, filename, code):
+        self.filename = os.path.normcase(os.path.abspath(filename))
+        with open(self.filename, "w") as f:
+            f.write(code)
```
alexmojaki, line 69: This should really happen in python_runner, I've noticed not having a real file causing bugs in futurecoder.

What's the normcase for?

src/workers/python/init.py
```diff
@@ -39,10 +101,19 @@ def show():
 
         matplotlib.pyplot.show = show
 
+    def pre_run(self, source_code, mode="exec", top_level_await=False):
+        self.override_globals()
+        return super().pre_run(source_code, mode=mode, top_level_await=top_level_await)
+
+    def execute(self, code_obj, source_code, mode=None):  # noqa
+        return eval(code_obj, self.console.locals)  # noqa
```
alexmojaki, line 109: What's this about?

src/workers/python/init.py
winniederidder: The normcase is what Pdb uses to resolve filenames, so I figured I would use their method to construct filenames so that every module would find and use the same file.

src/workers/python/init.py
winniederidder: PythonRunner only overrides methods when execute is called. I moved that to override_builtins to ensure they are overwritten early enough so that everything is always captured. I needed it for something Pdb-related, but I also am not sure why python_runner only applies the patches at execute time (and thus every time code is executed).

src/workers/python/init.py
winniederidder: I'm still unsure what the best approach is. The makeRunnerCallback method expects an array of callbacks, which becomes quite unwieldy. I moved to a PubSub system to prevent needing to pass many callbacks. I would think that adding a catch-all to makeRunnerCallback would make it a bit more user-friendly for this specific case, but I wouldn't call this part of my code stable yet.

src/workers/python/init.py
```diff
@@ -72,6 +75,7 @@ def override_globals(self):
         os.environ["MPLBACKEND"] = "AGG"
         sys.stdin.readline = self.readline
         builtins.input = self.input
+        time.sleep = self.sleep
```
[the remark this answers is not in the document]
winniederidder: The patches I discussed in a different comment.
For the recursion limit, I wasn't sure if you already set it somewhere. pyodide_worker_runner was freshly integrated so I didn't know this was the case.

src/util/Util.ts
winniederidder: Strings with a content type kind of keep things manageable. Pyodide has proper conversion methods, but other programming environments might not.  The current data type in BackendEvent for "data" is string, so I stuck to that for now instead of adding numbers to the mix.

src/workers/python/init.py
winniederidder: Is this then a feature you will add to python_runner, or what is our plan with this? The current python_runner API always works with source code or code objects, which is the reason why I added this myself.

src/workers/python/init.py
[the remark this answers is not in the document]
alexmojaki: The reason I ask is that you have parameters with defaults as though you want it sometimes change it.

src/workers/python/init.py
alexmojaki: I think this exact code should be in pre_run next to the linecache stuff, but with errors caught in case python_runner is used in places where writing to the filesystem is not an option.

src/CodeRunner.ts
```diff
@@ -119,9 +119,8 @@ export class CodeRunner {
             type: BackendEventType.End,
             data: "User cancelled run", contentType: "text/plain"
         });
-        // Since we use workers, the old one must be entirely replaced to interrupt it
-        this.backend.interrupt();
-        return this.start();
+        await this.backend.interrupt();
+        this.setState(RunState.Ready);
```
alexmojaki, line 123: Interrupting no longer necessarily means that the program will stop, because KeyboardInterrupt can be caught. pdb in particular does that. You need to catch `InterruptError` where `client.call` is called, and that will indicate that the worker was terminated because other interruption methods weren't available.

src/workers/python/python_package.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -eux
+
+rm -rf ./package ./package.tar
+mkdir ./package
+pip install -t ./package python-runner friendly_traceback jedi
```
alexmojaki, line 7: You may want to start pinning dependency versions, especially for friendly_traceback

src/workers/python/init.py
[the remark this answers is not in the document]
winniederidder: Maybe users of the package run into issues and want to increase/decrease it, so it's flexibility at very low cost (although I indeed do not see immediate reason to change it).

src/workers/python/python_package.sh
```diff
@@ -0,0 +1,10 @@
...
+cp -r ./papyros ./package
+cd ./package
+tar cfv ./../package.tar *
```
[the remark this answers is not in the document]
winniederidder: I believe npm compresses the bundle, so I don't think I need to add something here. Otherwise the webpack loader also needs a decompressing step, which might not be supported.

src/workers/python/papyros/autocomplete.py
```diff
@@ -0,0 +1,40 @@
...
+        import jedi
+        s = jedi.Script(context["text"])
+        # Convert Jedi completions to CodeMirror objects
+        options = [convert_completion(c, i)
+                   for (i, c) in enumerate(s.complete(line=context["line"], column=context["column"]))]
+        if "." in before["text"]:
+            complete_from = before["to"]
+    results = dict(options=json.dumps(options), contentType="text/json")
+    results["from"] = complete_from
+    return to_js(results)
```
alexmojaki, line 40:
```suggestion
    return dict(options=options, from=complete_from)
```

You don't need to_js here since you're already using convert on the other side. That may be while you still had a Map instead of an object. And you don't need to dump JSON when you're going to parse it immediately.

src/workers/python/PythonWorker.worker.ts
```diff
@@ -78,8 +78,10 @@ class PythonWorker extends Backend {
     override async autocomplete(context: WorkerAutocompleteContext):
         Promise<CompletionResult | null> {
         // Do not await as not strictly required to compute autocompletions
-        this.pyodide.loadPackagesFromImports(context.text);
-        const result = this.convert(await this.papyros.autocomplete(context));
+        const result = (await Promise.all([
+            this.pyodide.pyimport("pyodide_worker_runner").install_imports(context.text),
+            this.convert(await this.papyros.autocomplete(context))
```
alexmojaki, line 83: If you await this here then it's no longer a promise and the Promise.all is redundant.

winniederidder: Indeed, the await in line 83 is in the wrong location.

src/workers/python/init.py
winniederidder: Should I make a PR for those patches (together with the file-creation) in the python_repository or is this something you want to do yourself?

src/workers/python/papyros/autocomplete.py
winniederidder: This was indeed one of the reasons maps popped up. I removed both occurences of to_js for this. Side note, from is a Python keyword, which is why I can't use the kwargs-constructor and need to set it using ["from"]

src/workers/python/init.py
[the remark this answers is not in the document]
alexmojaki: Hmmmm, but users might want to change it in a script, and not be happy about it resetting each time.

src/workers/python/init.py
alexmojaki: I'd appreciante a PR :)

Maybe PyodideRunner could also optionally `install_imports` in pre_run/execute.

src/workers/python/papyros/autocomplete.py
alexmojaki: Ah, I forgot about the keyword issue. But you could use a dict literal.

src/workers/python/init.py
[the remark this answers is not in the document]
winniederidder: The limit can be set by altering the value for Papyros.limit, so this shouldn't pose an issue.

src/workers/python/init.py
winniederidder: I'll move the overriding there as well, and allow users to toggle it? (as discussed above) I'll await another collaborator invite then?

src/workers/python/init.py
winniederidder: Here, I meant that makeRunnerCallback expects an object with at least 2 callbacks and then extra callbacks per possible type. I have an error callback, the interrupt just in case, and potentially a debugging callback which makes this unwieldy to pass around. If more types get added, this also poses issues for makeRunnerCallback. The translation in Python is not exactly ideal either, so I'll think about a hybrid approach to clean this up.

src/workers/python/papyros/papyros.py
```diff
@@ -101,6 +101,13 @@ def show():
 
         matplotlib.pyplot.show = show
 
+    async def install_imports(self, source_code, ignore_missing=True):
+        try:
+            await install_imports(source_code)
+        except ValueError:
```
alexmojaki, line 107: When is this raised?

winniederidder: Mainly during autocompletion: when users are still typing and the autocomplete method is called, I try to install the imports   from what is currently present. If they are typing an import call, (e.g. import mat) when going for import math, this leads to trying to import the module 'mat', which leads to a ValueError.

src/CodeRunner.ts
winniederidder: By default, calling interrupt on a PyodideClient leads to a PythonError insteads of an InterruptError as the internal KeyBoardError is raised within Pyodide. Is your expected way of handling this to catch this in run_async?  Don't we want to throw an InterruptError in SyncClient.call by default, or should I use a custom interrupter for this?

alexmojaki: The user's code might catch `KeyboardInterrupt` (a notable case is pdb) and there's no way for the client to know whether that will happen. It mustn't throw InterruptError because the worker may not be done processing and ready for more.

Letting `KeyboardInterrupt` bubble up from python_runner's _execute_context without going through serialize_traceback is useful in futurecoder where significant work usually happens after running code to check if the answer was correct for the current step. But for the sake of generality, I suggest removing that case from _execute_context, and futurecoder can add a bare `raise` in serialize_traceback.

src/workers/python/init.py
winniederidder: > 
> Maybe PyodideRunner could also optionally `install_imports` in pre_run/execute.

This installation is async, so pre_run or execute is not possible. Should I add it to run_async instead?

winniederidder: Making other methods async might cause trouble in your other packages. However, if we are making such changes, we might want to make the source_code also a member variable, so it doesn't need to be passed along to every method?

alexmojaki: I forgot about the asyncness. Forget that part.

winniederidder: Having this method in python_runner might be nice, but i'll keep it within Papyros for now then?

src/workers/python/papyros/papyros.py
```diff
@@ -112,41 +98,25 @@ def pre_run(self, source_code, mode="exec", top_level_await=False):
...
+            try:
+                await self.install_imports(source_code, ignore_missing=False)
+                code_obj = self.pre_run(source_code, mode=mode, top_level_await=top_level_await)
+                if code_obj:
+                    result = self.execute(code_obj, mode)
+                    while isinstance(result, Awaitable):
+                        result = await result
+                    return result
+            except KeyboardInterrupt:
+                self.callback("interrupt", data="KeyBoardInterrupt", contentType="text/plain")
```
alexmojaki, line 112: Are you actually using this? At this point you already know the program is ending. Either way, if you handle this in serialize_traceback then you should be able to avoid copying run_async.

src/workers/python/papyros/papyros.py
```diff
@@ -112,41 +98,25 @@ def pre_run(self, source_code, mode="exec", top_level_await=False):
...
-                    # handle the exception
-                    raise
-        except KeyboardInterrupt:
-            self.callback("interrupt", data="KeyboardInterrupt", contentType="text/plain")
-
-    def serialize_syntax_error(self, exc, source_code):
+    async def run_async(self, source_code, mode="exec", top_level_await=True):
+        with self._execute_context():
+            try:
+                await self.install_imports(source_code, ignore_missing=False)
```
alexmojaki, line 104: Since the result from `execute` is already awaited, maybe you could override it with an async method and put this in there to avoid copying run_async?

src/workers/python/papyros/papyros.py
winniederidder: The KeyboardInterrupt is captured in _execute_context via
```python
except BaseException as e:
    self.output("traceback", **self.serialize_traceback(e))
```
I could also check in the event translation part for a KeyboardInterrupt error name, but output expects strings (which is why my FriendlyErrors are still dumped to strings using json) so the check would not really be clean (would need json parsing to cover all edge cases). If you have a possible fix for this, then I will gladly incorporate it.

src/workers/python/papyros/papyros.py
winniederidder: Is overriding a non-async method with an async one considered good practica? Or do you expect me to return a coroutine object instead of changing the method definition?

src/workers/python/papyros/papyros.py
alexmojaki: I meant you could just do this in `serialize_traceback`:

```python
if isinstance(e, KeyboardInterrupt):
    self.callback("interrupt", ...)
    # maybe return reduced output here
```

assuming that event type is valuable somehow.

> output expects strings (which is why my FriendlyErrors are still dumped to strings using json)

You could set text to `""` or a simple formatted traceback, then return the actual data in a dict without dumping to json.

The output system could be a bit less opinionated and not force you to supply `text`. `OutputBuffer` concatenates `text` values where possible, but realistically that only happens with stdout/stderr and maybe custom output types, not `traceback`. futurecoder concatenates the text from all parts into a single string which is useful for a few things, so it's nice when `text` always contains something human readable, but that's futurecoder.

src/workers/python/papyros/papyros.py
alexmojaki: I'm guessing it'd be OK to override with async.

src/workers/python/python_package.sh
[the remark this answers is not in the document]
winniederidder: Have you tried this compressing yourself? I have tried different variations to compress the tar into a tar.gz, using python or something like tar -czf, but I keep getting the following error:
```
comlink.mjs:69 Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "/lib/python3.9/shutil.py", line 1184, in _unpack_tarfile
    tarobj = tarfile.open(filename)
  File "/lib/python3.9/tarfile.py", line 1616, in open
    raise ReadError("file could not be opened successfully")
tarfile.ReadError: file could not be opened successfully

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/lib/python3.9/site-packages/pyodide/_util.py", line 33, in unpack_buffer_archive
    shutil.unpack_archive(f.name, extract_dir, format)
  File "/lib/python3.9/shutil.py", line 1247, in unpack_archive
    func(filename, extract_dir, **dict(format_info[2]))
  File "/lib/python3.9/shutil.py", line 1186, in _unpack_tarfile
    raise ReadError(
shutil.ReadError: /tmp/tmpm0qlluh_ is not a compressed or uncompressed tar file

    at new_error (pyodide.asm.js:14:247183)
    at pyodide.asm.wasm
    at pyodide.asm.wasm
    at Object.Module._pythonexc2js (pyodide.asm.js:14:944230)
    at Object.Module.callPyObjectKwargs (pyodide.js:1:10739)
    at Proxy.callKwargs (pyodide.js:1:18742)
    at Object.unpackArchive (pyodide.js:1:24012)
    at eval (PythonWorker.worker.ts?./node_modules/ts-loader/index.js:76:29)
    at step (PythonWorker.worker.ts?./node_modules/ts-loader/index.js:56:23)
    at Object.eval [as next] (PythonWorker.worker.ts?./node_modules/ts-loader/index.js:37:53)
```
Manually calling shutil.unpack_archive gives me the files, so something must be going wrong when using url-loader to obtain the tar.gz contents. Does this sound familiar to you, or do you have a similar experience?

src/CodeRunner.ts
```diff
@@ -0,0 +1,277 @@
...
+     * Interrupt the currently running code
+     * @return {Promise<void>} Promise of stopping
+     */
+    async stop(): Promise<void> {
+        this.setState(RunState.Stopping);
+        BackendManager.publish({
+            type: BackendEventType.End,
+            data: "User cancelled run", contentType: "text/plain"
+        });
+        await this.backend.then(b => b.interrupt());
```
alexmojaki, line 128: This is still assuming that nothing will catch the KeyboardInterrupt. You can only do this if `call` throws `InterruptError`. Treat client.interrupt as a suggestion/request.

src/CodeRunner.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    /**
+     * Set the used programming language to the given one to allow editing and running code
+     * @param {ProgrammingLanguage} programmingLanguage The language to use
+     */
+    async setProgrammingLanguage(programmingLanguage: ProgrammingLanguage): Promise<void> {
+        if (this.programmingLanguage !== programmingLanguage) { // Expensive, so ensure it is needed
+            await this.backend.then(b => b.interrupt());
```
alexmojaki, line 137: If you want to be sure here then use client.terminate(), but then the client must be deleted and not reused.

src/CodeRunner.ts
```diff
@@ -0,0 +1,277 @@
...
+            if (this.state !== RunState.Stopping) { // Was interrupted
+                BackendManager.publish({
+                    type: BackendEventType.End,
+                    data: endMessage, contentType: "text/plain"
+                });
+            }
+            const end = new Date().getTime();
+            this.setState(RunState.Ready, t("Papyros.finished", { time: (end - start) / 1000 }));
+            if (terminated) {
+                await this.start();
```
alexmojaki, line 273: comsync already restarts the worker after terminating. Also, you can end up here if you interrupt/terminate the client for the sake of switching programming language.

src/workers/javascript/JavaScriptWorker.worker.ts
```diff
@@ -121,7 +120,8 @@ class JavaScriptWorker extends Backend {
         return ret;
     }
 
-    override _runCodeInternal(code: string): Promise<any> {
+    override runCode(extras: SyncExtras, code: string): Promise<any> {
+        this.extras = extras;
```
alexmojaki, line 124: I assume that input is still supported in JS. What about sleeping?

src/workers/python/papyros/papyros.py
```diff
@@ -0,0 +1,164 @@
...
+            if event_type == "output":
+                for part in data["parts"]:
+                    typ = part["type"]
+                    if typ in ["stderr", "traceback", "syntax_error"]:
+                        cb("error", part["text"], contentType=part.get("contentType"))
+                    elif typ in ["input", "input_prompt"]:
+                        # Do not display values entered by user for input
+                        continue
+                    else:
+                        cb("output", part["text"], contentType=part.get("contentType"))
```
alexmojaki, line 50: I think you should embrace the python_runner output part types approach. You're sending stderr and stdout to separate event types and code paths, but they're almost exactly the same thing, just different colours. Meanwhile you're combining stderr and tracebacks which get handled completely differently. Ultimately these are all just different types of output which all get rendered to the same output area, even if it's in different ways.

src/workers/python/init.py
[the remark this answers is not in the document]
alexmojaki: resolved

src/workers/python/init.py
alexmojaki: It sounds like it would be better if makeRunnerCallback didn't try to be so helpful with other event types and just let you provide a function to act as a default. That way you could just let other events pass through to your own system.

src/workers/python/init.py
alexmojaki: resolved

src/workers/python/papyros/papyros.py
alexmojaki: resolved

src/workers/javascript/JavaScriptWorker.worker.ts
winniederidder: Sleeping doesn't really seem to be a thing in JS. What exists is setTimeout, which is non-blocking. In order to properly support that, I think I would have to store the timers internally and await them before giving control back to the user. Our focus is currently on Python, so I don't plan on changing this yet.

src/workers/python/init.py
winniederidder: A catch-all would indeed be more practical in my opinion.

src/workers/javascript/JavaScriptWorker.worker.ts
alexmojaki: > What exists is setTimeout, which is non-blocking. In order to properly support that, I think I would have to store the timers internally and await them before giving control back to the user.

What exactly are you saying? Using setTimeout as usual doesn't work properly? Or do you mean making setTimeout blocking?

You could add a custom `sleep` function using `extras.syncSleep()`, which currently just means running the callback with the sleep event.

winniederidder: What I meant is that a blocking sleep call does not exist in JavaScript by default. It can be emulated with Promises and awaiting them, but I haven't played around a lot with async/await in the JavaScript runner.
setTimeout is non-blocking and works, but these calls are not "awaited", so if you write eg 
```
setTimeout(() => console.log("Timer"), 5000)
```  
then the run will be finished immediately and the log statement will appear later in the actual console instead of the output field. This is because of the global overrides being done at the start and being cleared at the end of a run. That's why some awaiting mechanism should be added for these timeouts.

src/workers/python/papyros/papyros.py
winniederidder: These things are a bit too tied to Python, so they're not easily generalised to other languages (e.g. due to presence of snake_case). With the eventing system, I think it's nice to be able to only listen to error messages instead of getting all output. Finally, I don't think our target group will often write to stderr, so the focus is on the tracebacks as they will be the main reason for error displaying. The current translations seem fine to me.

src/CodeRunner.ts
winniederidder: The goal is to fully interrupt. Ideally, some method should exist in synClient to force-interrupt (maybe a configurable timeout that terminates the worker if the interrupt was not successful?) Having to interpret client.interrupt as a suggestion is counter-intuitive to me.

src/CodeRunner.ts
winniederidder: Restarting the worker is not enough, as the launch method must also be called. This is done in start, so that should be fine. For the second case, I don't think anything will go wrong if start is called twice.

bmesuere (approved):

PR #114, opened 2022-04-06
Support linting

### LINKED ISSUE #42: Support linting
As we have a runtime environment (Python/JavaScript), we might also support linting while code is edited.

winniederidder:
Adds support for linting Python files using pylint with the config used in the Pythia judge.

Example output:
linting

Closes #42


alexmojaki:
Looks like the next release of Pyodide should support Pylint now that https://github.com/pyodide/pyodide/pull/2320 is merged.

But it also looks like lazy-object-proxy doesn't *need* the C extension code and it can fallback to pure-python (see https://github.com/ionelmc/python-lazy-object-proxy/blob/master/src/lazy_object_proxy/__init__.py) which is how it supports pypy. So I expect it's already possible to use pylint now, just not via micropip.

bmesuere (approved):
Looks ok but, if possible, I would prefer pylint since we also use that in the python judge

winniederidder:
> Looks ok but, if possible, I would prefer pylint since we also use that in the python judge

By using the tar and pre-installing, it seems to be possible to use Pylint (after a minor amount of patching to get it to work within Pyodide). Which linting rules are used in the python judge? It should be possible to inform Papyros about these rules, so that the warnings are consistent.

alexmojaki:
Pyodide 0.20.0 has been released, it should allow installing pylint normally.

winniederidder:
> Pyodide 0.20.0 has been released, it should allow installing pylint normally.

I am doing that, but it required a minor patch around os.devnull to be able to import it. I haven't tried it in v0.20 yet as that as we planned a PR for that on pyodide-worker-runner.

bmesuere (approved):

PR #116, opened 2022-04-14
Support snoop

winniederidder:
Adds support for Backends to add more run modes (examples can be exec, snoop, birdseye, debug, etc)
Implements the snoop-mode for Python, showing the user a detailed trace of which parts their code visited.
Example output:
snoop

The current layout is quite basic, I'm not sure what style to apply. Maybe a little lighter in color to show it's not regular output  (the indent already makes it quite clear)


alexmojaki:
In futurecoder I use https://github.com/drudru/ansi_up to interpret the ANSI colour codes produced by snoop. For syntax highlighting you could pick a theme from http://help.farbox.com/pygments.html

winniederidder:
Example of possible Snoop output:
snoop

alexmojaki:
You really need a fixed width font there. Also I suggest setting the filename to `/my_program.py`.

src/OutputManager.ts
```diff
@@ -81,7 +85,13 @@ export class OutputManager {
...
                 .join("\n");
         }
-        return `<span class="${className}">${escapeHTML(spanText)}</span>`;
+        // If the HTML can be interpreted as colored, then render it that way
+        const coloredText = this.ansiUp.ansi_to_html(spanText);
+        if (coloredText !== spanText) {
+            return coloredText;
+        } else {
+            return (`<span class="${className}">${escapeHTML(spanText)}</span>`);
+        }
```
alexmojaki, line 94:
```suggestion
        return `<span class="${className}">${this.ansiUp.ansi_to_html(spanText)}</span>`;
```

winniederidder: Does ansiUp apply escaping to text that it doesn't modify?

winniederidder:
> You really need a fixed width font there. Also I suggest setting the filename to `/my_program.py`.

The font css was being applied incorrectly, so that's fixed now. The filename is indeed an issue. The problem is to let all libraries find the source file and still displaying it in a nice fashion. I'm not entirely sure how to get that working consistently.

src/OutputManager.ts
winniederidder: Seems to be the case, so I'll remove the if

alexmojaki:
> The problem is to let all libraries find the source file and still displaying it in a nice fashion. I'm not entirely sure how to get that working consistently.

Can you explain more? Which libraries are you talking about? What's wrong with `filename="/my_program.py"` on the python runner? That's what I'm doing.

winniederidder:
> Can you explain more? Which libraries are you talking about? What's wrong with `filename="/my_program.py"` on the python runner? That's what I'm doing.

I'm talking about a recent addition to python_runner, being
```python
def set_filename(self, filename):
    self.filename = os.path.normcase(os.path.abspath(filename))
```
which we did as e.g. Pdb uses this to resolve filenames, whereas other libraries just use relative paths or something similar. So the filename stored is an absolute path, causing it to display like this in other areas too.

alexmojaki:
Right, but I'm saying the actual absolute path can be `/my_program.py`, at the root.

winniederidder:
> Right, but I'm saying the actual absolute path can be `/my_program.py`, at the root.

Oh I see, thanks for this!

alexmojaki:
What happened to this? :cry:

winniederidder:
> What happened to this? 😢

While this PR is functional, we decided to postpone providing debuggers in Papyros to ensure we offer functionality that users can and will use. A separate thesis might cover this in the futrue, at which point this PR might be bought back to life.

PR #125, opened 2022-04-19
Always catch KeyboardInterrupt

winniederidder:


winniederidder:
> Probably better to catch `BaseException` which covers everything.

I believe the key issue was that KeyboardInterrupt is not a BaseException, so that's why KeyboardInterrupt must be caught specifically. I'm not sure which other important Exceptions would be Base but not Exception.

alexmojaki:
BaseException is the superclass of all these things. Exception is slightly narrower. KeyboardInterrupt is a BaseException but not an Exception.

winniederidder:
> BaseException is the superclass of all these things. Exception is a slightly narrower subclass. KeyboardInterrupt is a BaseException but not an Exception.

It is not.
keyboardinterrupt
.

winniederidder:
> Your arguments are backwards, and isinstance doesn't work like that on class objects.
> 
> ```
> >>> issubclass(KeyboardInterrupt, BaseException)
> True
> >>> issubclass(Exception, BaseException)
> True
> ```

I stand corrected and have changed it to BaseException.

PR #127, opened 2022-04-19
Update favicon and dependencies

winniederidder:
Noticed during the demo that the default React favicon was still being used.
This PR changes it to the Github scroll emoji.
favicon


winniederidder:
Updated yarn.lock to also close #113, #115 and #119

bmesuere (approved):
I would from now on use the dependabot PRs to update the dependencies to better track the changes and potential problems.

Issue #76, opened 2021-12-12
Refactor input handling


winniederidder:
- [x]  If statements based on the Enum value are prevalent. This should be replaced by classes to abstract away this logic.
- [ ] Add buttons to start/stop buttons with an inline svg
- [x] Issue #72 
- [x] Issue #75

Issue #72, opened 2021-12-09
Automatically scroll down in output textarea upon overflow


pdawyndt:
As output textarea overflows with new output, make it automatically scroll to the newly appended lines

PR #134, opened 2022-04-28
Update input and output handling

winniederidder:
This PR updates a couple of things to improve how input and output is handled.
When rendering new output, the area scrolls to the bottom, closing #72 
~~When starting code in batch mode, the input is sent up front. (part of #76 )~~
The potential output is capped to prevent message passing issues in Comlink. (closing #17 )
Also fixes a minor translation issue where a Dutch key was missing (switch_input_mode_to)


bmesuere:
A user suggested to display a message if the code executed without output. I agree this might be a useful addition. We must of course make sure that it is clear that that message is not output itself. It might be an option to update the placeholder from "output will appear here" to something like "code executed without generating output"

alexmojaki:
> When starting code in batch mode, the input is sent up front.

I'm curious what difference this makes since the comsync issues was solved? Is it a workaround for deployments that can't support a sync-message channel?

alexmojaki:
> It is marginally faster

Testing on https://papyros.dodona.be/, looks like reading `input()` 100 times takes about 0.1 seconds.

alexmojaki:
> In theory, it should be faster when using a service worker

What makes you say that?

In any case, I just wanted to make 'marginally faster' precise and concrete for everyone. Unless you need to read thousands of lines of input in Papyros, speed seems to be a non-issue.

winniederidder:
> > In theory, it should be faster when using a service worker
> 
> What makes you say that?
> 
> In any case, I just wanted to make 'marginally faster' precise and concrete for everyone. Unless you need to read thousands of lines of input in Papyros, speed seems to be a non-issue.

There is no need for two network requests that contact the service worker per line, so I would be very surprised if it could be slower. Service worker requests are likely quite cheap and indeed, the performance benefit will be negligible. I am still awaiting feedback from my peers on whether we want to keep this.

bmesuere (commented):
Won't sending all input at once cause issues when the input is modified after clicking run?

winniederidder:
> Won't sending all input at once cause issues when the input is modified after clicking run?

This will cause inconsistencies if the code is slow and the user manages to change the input before it is consumed.
We could migrate this code to a separate InputMode that would serve as a fallback in case the service worker fails, where input is then also not modifiable during execution.

bmesuere:
This then seems to add unnecessary complexity for minimal gains (unless I'm missing something). We can discuss this tomorrow during the meeting.

bmesuere (approved):

Issue #133, opened 2022-04-27
Maintainability of Papyros


winniederidder:
This issue contains the tasks that must be done to ensure the package attains a maintainable state. 
- [x] Add tests to the package 
- [x] Configure Dependabot for version management and upgrade packages to the latest version
- [x] Finalize and document the public API (which items are exported and which functions do they provide)
- [x] Work with changelog releases
- [x] Release version 1.0.0 
- [x] Provide a separate GitHub-pages URL for the main version and the develop version

PR #135, opened 2022-04-29
Main and dev GitHub-pages

winniederidder:
This PR provides support for deploying two versions of Papyros: the main branch that aligns with the current release, and a dev version to try things out in the browser.
Two URLs are now supported:
main version on https://papyros.dodona.be
dev version (current branch): https://papyros.dodona.be/dev/  (the trailing slash is imported to be routed to index.html, unclear why no automatic redirect is done)
Finishes a task of #133

### DIFF
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -15,7 +15,7 @@ jobs:
       name: production
       url: https://docs.dodona.be/papyros/
     steps:
-      - name: Checkout
+      - name: Checkout branch
         uses: actions/checkout@v2.3.1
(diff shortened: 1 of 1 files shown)


bmesuere (approved):
In the service worker case, won't 2 service workers on the same domain cause issues?

winniederidder:
> In the service worker case, won't 2 service workers on the same domain cause issues?

Their domain should not be the same as the url is now /dev/serviceWorker vs /serviceWorker. According to https://w3c.github.io/ServiceWorker/#service-worker-registration-concept, the most specific service worker will be used, so they should not interfere.

PR #174, opened 2022-05-10
Improve linting and alleviate double display issue

### LINKED ISSUE #162: Avoid double explanation of linting messages
I can see why we have a double annotation of linting messages, indicating both the line (left margin) and column (inline triangle) where the issue was found. However, we should avoid showing the explanation twice while hovering over the visual indicators.

image

winniederidder:
This PR upgrades the linting provided by Papyros. Newer versions of Pylint provide more accurate issue descriptions, allowing more accurate indications in the editor. For example, the end line and end column to properly mark the statement. The lint gutter is now also on the left, which should help to prevent double displays of issues (as the built-in filter features of CodeMirror are not suitable for this cause).
Example output:

linting-new

Closes #162


bmesuere (approved):
👍

PR #177, opened 2022-05-12
feat: Save code and restore after reload

### LINKED ISSUE #89: Keep code after page reload
Should avoid that (inexerienced or experienced whatsoever) programmers loose their code after a page reload.

winniederidder:
This PR updates the stand-alone mode of Papyros to store the user's code every so often in local storage, after which it will be restored when the page is reloaded.

Closes #89


bmesuere (commented):
Can't we be more clever about this? E.g. by only saving the code when the beforeunload event of the browser fires?

alexmojaki:
FWIW futurecoder now saves most of the redux state (including the editor content) locally after every change, so every keystroke in the editor triggers an immediate save in the background, and it seems to do it very quickly with no visible downsides. It uses `localForage` so I think typically that means IndexedDB, not sure if that makes a difference.

bmesuere:
To be clear, onchange is also ok for me (I think codemirror has a few events you can hook into).
I wouldn't use localForage since it's an extra dependency and localStorage is perfectly fine for this.

bmesuere:
If it's possible to hook into a codemirror event, this would probably be the best?

winniederidder:
> If it's possible to hook into a codemirror event, this would probably be the best?

I added an onChange-option to the CodeEditor to allow processing of the new contents.

bmesuere (approved):

PR #178, opened 2022-05-12
feat: Better automatic imports

### LINKED ISSUE #87: Support plotting in Safari
Also give Safari users superpower to run code like this in Papyros:

image

### LINKED ISSUE #98: Show modules being loaded in statusbar

winniederidder:
This PR adds support for two things.
First off, hidden dependencies are automatically installed and imported if possible.
Secondly, to make this more obvious, support was added to show that loading is taking place.
loading-modules
On a side note, the installing of packages takes a toll on the Astroid-cache used in Pylint. To alleviate this, the cache is no longer cleared there and instead a temporary file is used to prevent caching issues from occurring. This prevents a huge stream of warnings from Pylint from popping up during installations.

Closes #98 
Closes #87 by automatically importing hidden dependencies


src/workers/python/papyros/papyros.py
```diff
@@ -124,6 +124,17 @@ async def run_async(self, source_code, mode="exec", top_level_await=True):
...
+                # Try to automatically install missing dependencies
+                # As they sometimes might be hidden within libraries
+                try:
+                    self.callback("loading", data=dict(loading=True, modules=[mnf.name]), contentType="application/json")
+                    await self.install_imports(f"import {mnf.name}", ignore_missing=False)
+                    self.callback("loading", data=dict(loading=False, modules=[mnf.name]), contentType="application/json")
+                    return await self.run_async(source_code, mode=mode, top_level_await=top_level_await)
+                except:
+                    # If the module is truly not findable, raise the error again
+                    raise mnf
```
alexmojaki, line 137: This will translate regular exceptions in the user's code into ModuleNotFoundError even if the import installed just fine.

src/workers/python/PythonWorker.worker.ts
```diff
@@ -65,17 +65,44 @@ class PythonWorker extends Backend<PyodideExtras> {
...
      */
-    private async installImports(code: string, ignoreMissing: boolean): Promise<void> {
+    private async installImports(code: string): Promise<void> {
         if (this.installPromise == null) {
-            this.installPromise = this.papyros.install_imports.callKwargs({
-                source_code: code,
-                ignore_missing: ignoreMissing
-            });
+            this.installPromise = (this.pyodide as any)
+                .loadPackagesFromImports(code,
```
alexmojaki, line 104: Why this change?

winniederidder: This is specifically to have access to the message callback option provided by Pyodide. Doesn't Pyodide also perform micropip-installs if needed, or is that specific to pyodide-worker-runner?

alexmojaki: I don't think it does, see https://github.com/pyodide/pyodide/issues/1470#issuecomment-907815342

You can check by seeing what happens when you import something that isn't pyodide-specific. Right now it sort of looks like `loadPackagesFromImports` will miss it but the new system might still catch it at runtime.

winniederidder: Should I look into opening a PR on pyodide-worker-runner to accept the callbacks instead? Then we can have the best of both worlds.

winniederidder:
@bmesuere The latest version of pyodide-worker-runner now supports granular import notifications, so this PR is properly ready for review.

bmesuere (approved):

PR #179, opened 2022-05-12
feat: Add keybind to run code

### LINKED ISSUE #36: Key binding for running the code
Investigate if it would be useful to have a shortkey for running the code.

winniederidder:
This PR changes some keybinds.
Ctrl-Enter can now be used to run the code without needing to click the button
The original keybinding (insertBlankLine) is now assigned to Shift-Enter, similar to how other editors handle this (like the GitHub comment window here)

Closes #36

### DIFF
--- a/src/CodeEditor.ts
+++ b/src/CodeEditor.ts
@@ -6,7 +6,7 @@ import {
     CompletionSource, autocompletion,
     closeBrackets, closeBracketsKeymap, completionKeymap
 } from "@codemirror/autocomplete";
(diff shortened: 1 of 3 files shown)


bmesuere (approved):
Since you're hooking into the keybindings of codemirror, I assume that on mac cmd+enter will be used instead of ctrl (which is good)?

winniederidder:
> Since you're hooking into the keybindings of codemirror, I assume that on mac cmd+enter will be used instead of ctrl (which is good)?

This should be the case indeed, according to the details I found regarding this on CodeMirror. (where Mod should be Ctrl on non-Mac and Cmd on Mac).

PR #194, opened 2022-05-21
Scope tailwind

winniederidder:
This PR limits the css provided by Tailwind to only the parts we actually use. This was done by filtering the preflight.css . These have been manually included in the Papyros.css files, as Papyros itself still depends on those to be displayed correctly.
Tailwind does apply its prefix to the utilities, so these do not clash with other user's css.

Closes https://github.com/dodona-edu/dodona/issues/3638


bmesuere (changes requested):
Please add a comment with a reference to the file where these styles were copied from.

bmesuere (approved):

PR #195, opened 2022-05-22
feat: support for run modes and doctests

### LINKED ISSUE #183: Detect doctests and allow running them
When programming in Python, doctests are often used to specify tests so the programmer can quickly check if everything is working as intended. We could detect that these tests are present and present the user with a "Doctest"-button to quickly run these without needing a main-block. We could then also show a message when every tests succeeds, as by default no output is generated which could cause confusion.

winniederidder:
This PR extends Papyros with the following features:
ChangeListeners to the user's code can now specify a delay before they are called, i.e. delay ms will pass after the last change to the document before they are called. This allows running more expensive checks for advanced functionality with minimal impact (similar to how linting is done)
Backends can provide extra running modes depending on the contents of the editor. Implementation has been provided for a doctest-runner: enabled when doctests are detected and automatically runs them. Currently, the run happens in verbose mode to make it clear what happened to the user (preventing empty outputs indicating success).
Example usage:
With failures -
doctests
Without failures -
doctest-success

Closes #183


winniederidder:
Closed in favor of #216 as somehow two PRs got merged on accident.

PR #216, opened 2022-05-30
feat: Run python doctests

### LINKED ISSUE #183: Detect doctests and allow running them

winniederidder:
This PR extends Papyros with the following features:
ChangeListeners to the user's code can now specify a delay before they are called, i.e. delay ms will pass after the last change to the document before they are called. This allows running more expensive checks for advanced functionality with minimal impact (similar to how linting is done)
Backends can provide extra running modes depending on the contents of the editor. Implementation has been provided for a doctest-runner: enabled when doctests are detected and automatically runs them. Currently, the run happens in verbose mode to make it clear what happened to the user (preventing empty outputs indicating success).
The run and stop button have also been combined into a single button as one was always disabled anyway.
Example usage:
papyros-doctests

Closes #183


winniederidder:
> * I would make this a neutral button instead of a green one (similar to the copy code to dodona)

Saved this as neutral-bg in the tailwind config.

> * Can you add a doctest example to the examples section?
This example already exists, it's called "Doctests" and features code based on an implementation of factorial.

pdawyndt:
Now that you toggle between the `Run` and `Stop` buttons, could you make sure they have the same width. Now they have a slightly deviating with, which gives some unpleasant "jumping" behaviour.

winniederidder:
> Now that you toggle between the `Run` and `Stop` buttons, could you make sure they have the same width. Now they have a slightly deviating with, which gives some unpleasant "jumping" behaviour.

I could add a min-width, but since these are translated strings there's no easy way around that. Any language that is added later might run into different problems when the button text is changed.

bmesuere (approved):
Sorry, I must have been suffering from male pattern blindness.

I agree with @pdawyndt , adding a min-width to the run/stop button would be welcome. They are the same in Dutch so no problem for now.

PR #218, opened 2022-05-30
Translate and style CodeMirror extensions

### LINKED ISSUE #207: Translate and style search & replace toolbar
No Dutch translation of "search and replace" toolbar (ctrl-F).

image

winniederidder:
This PR adds support for translating phrases used by CodeMirror. It also applies typical Papyros-styling to common elements generated by common extensions like search.
Example of the new search panel (open with Ctrl-F)
translate-codemirror

Closes #207
Depends on #217


pdawyndt (approved):
LGTM

bmesuere (changes requested):
I would capitalize all buttons since we already do that for the other buttons in papyros (and all other buttons on Dodona.

In addition, the style is only halfway there. There's a gradient background that is not present in the "neutral" copy-to-dodona button above.

bmesuere (approved):

PR #219, opened 2022-05-31
Better handling of import errors

winniederidder:
This PR improves the handling of import errors. In cases where the user is still typing the full name of the import, it might happen that the automatic import system is triggered. It then tries to install a module that does not exist, which causes the message "Installing module" to display until a run is started. With these changes, this message will no longer appear. Further handling is not required, as a ModuleNotFoundError will be raised by the code when running it later. There is no need to show a warning early on, as most likely the user wasn't done typing yet. Linting will also show an error when modules are not found.


winniederidder:
> Maybe related. If I'm running the matplotlib demo for the first time after a page load I see two messages:
> 
> 1. `Loading ...`
> 2. `Running`
> 
> Total time is about 5-6 seconds, with about 1-2 seconds for loading the libraries. This still makes that the running phase takes about 4-5 seconds for the first run, and less than a second for the second run. Thus, it seems that some part of "installing modules in the runtime environment" is not completely included in the phase where the `Loading ...` message is shown.
> 
> My guess is that there are actually two phases related to loading modules _i_) downloading the wheels/modules and **ii**) installing the modules in the runtime environment, and that only the first of these gets the message `Loading ...` where the second is included in the phase where the `Running` message is displayed. I would find it more intuitive if the second phase of installing is excluded from the `Running` phase, either by including both installation phases in the same `Loading ...` phase (may rename as `Installing ...`) or using two distinct messages for loading and installing.

While I agree, I am unsure if this is feasible. The messages are displayed after micropip has finished installing. What happens between that part and the actual run starting, is not entirely clear and is not explicitly accessible. A possible workaround would be using "Loading" or something similar instead of "Running" when the user clicks the button, and letting the backend send a message right before they start actually running the code. The pre-run code (the other steps in e.g. the PythonRunner pipeline) would thus be seen as loading instead of running. However, there is a big chance this wouldn't solve the issue, as the user's code is the part that contains the actual "import"-calls that are most likely causing the delay. So, to combat that, we could use importlib during the installation part to also import the library already. Then I would think the installing in the runtime environment would be accurately captured.

winniederidder:
> > Maybe related. If I'm running the matplotlib demo for the first time after a page load I see two messages:
> > 
> > 1. `Loading ...`
> > 2. `Running`
> > 
> > Total time is about 5-6 seconds, with about 1-2 seconds for loading the libraries. This still makes that the running phase takes about 4-5 seconds for the first run, and less than a second for the second run. Thus, it seems that some part of "installing modules in the runtime environment" is not completely included in the phase where the `Loading ...` message is shown.
> > My guess is that there are actually two phases related to loading modules _i_) downloading the wheels/modules and **ii**) installing the modules in the runtime environment, and that only the first of these gets the message `Loading ...` where the second is included in the phase where the `Running` message is displayed. I would find it more intuitive if the second phase of installing is excluded from the `Running` phase, either by including both installation phases in the same `Loading ...` phase (may rename as `Installing ...`) or using two distinct messages for loading and installing.
> 
> While I agree, I am unsure if this is feasible. The messages are displayed after micropip has finished installing. What happens between that part and the actual run starting, is not entirely clear and is not explicitly accessible. A possible workaround would be using "Loading" or something similar instead of "Running" when the user clicks the button, and letting the backend send a message right before they start actually running the code. The pre-run code (the other steps in e.g. the PythonRunner pipeline) would thus be seen as loading instead of running. However, there is a big chance this wouldn't solve the issue, as the user's code is the part that contains the actual "import"-calls that are most likely causing the delay. So, to combat that, we could use importlib during the installation part to also import the library already. Then I would think the installing in the runtime environment would be accurately captured.
> 
> I'll take a look at adding calls to importlib to see if this resolves the timing issue.

Some quick testing shows that importing is not the costly aspect. Wrapping the actual execution part with callbacks seems to be accurate, so I could use that to provide better timings.

bmesuere (approved):
This highly needed, I bumped into this issue a few times myself during testing.

PR #270, opened 2022-06-30
Use hidden instead of display none to keep layout

winniederidder:
Changes how elements are hidden: now they affect the layout but are invisible, to keep everything consistent when dynamically hiding elements.

Related to https://github.com/dodona-edu/dodona/pull/3684#issuecomment-1171573934


bmesuere (approved):
Thanks!

PR #536, opened 2023-10-06
Add json tracer as debug option

jorg-vr:
This pr introduces a pythontutor based debugger

Generating the debug trace in python is handled by [json-tracer](https://github.com/dodona-edu/json-tracer)
Visualizing these traces is handled by [trace-component](https://github.com/dodona-edu/trace-component)

The new class `Debugger` is responsible of rendering the trace and delegating all relevant events.

The code editor is updated to show a line indicator in the sidebar.
The input component adds layout based on if a line is already used in the current debug step.
The output component add an opacity to output which is not present yet for the current debug step.

There is also a hard limit of 10 000 execution steps added to help students who encounter infinite loops.

This pr also made some styling changes to make it easier to override styles for users of the library.
Peek 2023-12-22 15-17

- [ ] update readme


chvp (approved):
(Only looked at the code, couldn't immediately find the functionality.)

PR #558, opened 2023-11-24
Cleanup and mimize build size

jorg-vr:
I used the new webpack 5 features to separate the workers from the main package https://webpack.js.org/guides/web-workers/

The library is no longer bundled by webpack.
It provides the javascript compiled by tsc and the css compiled by tailwind.

Bundling and minimization should be done by the importer.
This allows optimization of dependencies on their end. (Eg otherwise all codemirror code was loaded twice in dodona)

I used file-loader instead of url-loader in order to not make the whole python zip inlined in javascript.

The scope of the serviceworker had to move to the root as some parts of the code are now served from /assets in dodona

To test the impact of these build changes I made a pr in dodona https://github.com/dodona-edu/dodona/pull/5204
As can be seen there, the changes here cause some required changes there.


jorg-vr:
Old  dodona package size:
image

New dodona package sizes:
image
The new js does not yet include the 2.4MB zipped python code, but it is already a major reduction

bmesuere (approved):

Issue #623, opened 2024-02-12
Add debug state for editor


jorg-vr:
> I encountered a small issue when running a simple input/output exercise (sum of 2 numbers): the input of the test was correctly loaded into the input area, but the lines were marked as checked from the start. 

this would also solve #611

Some other suggestions, probably for a later PR:
- If you click debug, wait for the traces to generate and then edit the code you are still debugging the old code. In addition, there is no good way to make the code arrow go away. I would enter a "debug mode" when clicking debug which restricts some features to make this more clear (show stop button, read only editor).
- The "code arrow" doesn't scroll into view when it's not visible. Maybe we can reuse the active line feature of the editor?

_Originally posted by @bmesuere in https://github.com/dodona-edu/dodona/pull/5336#pullrequestreview-1874365973_

jorg-vr:
closed by #629

Issue #611, opened 2024-01-26
Output generated while tracing debug steps should also be greyed out


pdawyndt:
This is the initial view after debugging is started:

![image](https://github.com/dodona-edu/papyros/assets/5736113/dd0d996e-285e-4c62-9359-4d71a90dfdf4)

I would expect the generated output to be greyed out when in step 1, as is the case when going to a future step and then returning to step 1:

![image](https://github.com/dodona-edu/papyros/assets/5736113/cfaf31bf-ab19-4c23-8c56-199f0a486cd6)

Actually, the arrow showing the line (statement) that will be executed next is also missing after initial launch of the debugger, and also the check marks of the consumed input are present. I guess this is all part of the same issue.

jorg-vr:
closed by #629

PR #629, opened 2024-02-13
Add clearer debug state for editor

### LINKED ISSUE #628: Awaiting coderunner.stop() should actually wait until it is stopped
see https://github.com/dodona-edu/dodona/pull/5362

jorg-vr:
This pr improves the debug state of the editor.

When a debug run is started, the editor becomes readonly.
All visualization of generated output, input used and active line is now based on the shown frame instead of the current phase of execution.
While the debug mode is active, the buttons to start a new execution are replaced by an end debug mode button.
Peek 2024-02-16 10-30

closes #628 #623 #611

### DIFF
--- a/src/Translations.js
+++ b/src/Translations.js
@@ -66,7 +66,10 @@ const ENGLISH_TRANSLATION = {
             "remove": "Remove"
         }
     },
-    "debug_mode_active": "Debug mode active",
+    "debug": {
+        "active": "Debug mode active",
+        "stop": "Stop debugging"
+    },
 };
 
 const DUTCH_TRANSLATION = {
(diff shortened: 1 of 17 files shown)


src/Translations.js
```diff
@@ -131,7 +134,10 @@ const DUTCH_TRANSLATION = {
             "edit": "Bewerk",
             "remove": "Verwijder"
         }
-    }
+    },
+    "debug": {
+        "stop": "Stop met debuggen"
```
bmesuere, line 139: `Stop debugger` is a little shorter

bmesuere (approved):
I left my functional comments in the other PR. Code-wise, this seems to make sense but it is very codemirror specific. Since I don't know that API, I can't really comment on this.

PR #878, opened 2025-09-18
Refactor papyros to webcomponents

### LINKED ISSUE #849: Lint output can refer to wrong lines
https://dodona.sentry.io/issues/1509182/?referrer=slack&notification_uuid=1a164c1c-c4b9-4b22-a559-f3f7e7c0401d&alert_rule_id=19800&alert_type=issue

jorg-vr:
This pr is a major refactor of papyros.

The main goal was to make the code more maintainable, by removing the custom rendering system.
The secondary goal was to replace tailwind by a more easily overwrite able styling system.

This pr splits all `view` functionality from all business and state logic.
I reorganized directories to make the new structure clearer
- `backend`: the code execution functionality, running in webworkers
- `frontend`: the code running in the main js thread
  - `state`: state management, eg is the debugger active
  - `components` : the visualization of this state
- `communication`: helper code to support frontend-backend communication

In this pr the `backend` and `communication` code are mostly unchanged.
The main change is the separation of the `frontend` in `state` and `components`.

## State
To access the state one requires an instance of `Papyros` as entrypoint.
By default a global instance `papyros` is exported, as in most cases you only need a single instance.

But this is set up in such a way that in theory you could create multiple Papyros instances to run multiple programs in the same environment. **Important note:** For this to be fully possible we the backend manager and the backend communication should also be refactored to support this. See future work.

As initializing the `backend` and `communication` is a heavy operation, this is not done on import, but should be done by explicitly calling `papyros.launch()`.

The state of the papyros instance has been split in multiple logical parts:
- `papyros.constants`: general settings and constants, mostly used for visualization, that can be overwritten
- `papyros.debugger`: keeps track of the debugframes and the actively shown debug frame
- `papyros.examples`: keeps track of all example code available
- `papyros.i18n`: all translations can be specified (and used) here
- `papyros.io`: keeps track of all input and output. You can subscribe to `awaitingInput` to `provideInput` when needed by the program
- `papyros.runner`: keeps track of code, execution state, programming language. Use `papyros.runner.start()` to start excecution
- `papyros.test`: keeps track of test code, appended to the end of the code document

#### Minimal setup without components
```ts
import {papyros} from "@dodona/papyros/Papyros";
papyros.launch();
papyros.runner.code = "print(input())";
papyros.io.subscribe(() => papyros.io.awaitingInput ? papyros.io.provideInput("foo"): "", "awaitingInput");
await papyros.runner.start();
console.log(papyros.runner.io.output);
```

## Components
The app consists of 4 main visual components. This are lit webcomponents, which all have only a single attribute that can be set: `papyros`. This expects a `Papyros` state instance. If non is given, the global `papyros` instance is used by default.

#### Minimal setup
```html
<script type="module">
 import {papyros} from "@dodona/papyros/Papyros";
 import "@dodona/papyros/components";
 papyros.launch();
</script>
<p-code-runner></p-code-runner>
<p-debugger></p-debugger>
<p-input></p-input>
<p-output></p-output>
```

### `<p-code-runner>`
This is a code mirror editor that allows editing, running and debugging the code papyros.
Additional buttons can be added through the `.buttons` slot, used here for the example picker.

image

### `<p-input>`
This is a component that lets you input either batch or interactive input, which it will pass to `papyros.io`

image
image

### `<p-output>`
Visualizes the output stream, combining std, stderr and images.
image

### `<p-debugger>`
Visualizes the debug traces using `@dodona/json-tracer`
image

## Theming
For theming I went fully into material design, using `@material/web` components  for buttons, sliders, inputs, etc..
I also used these material colors to create a theme for codemirror, and style all other elements.

This means that papyros can be completely customized by providing `--md-sys-color-...` css variables for you desired theme. Three example themes are provided in dark and light mode in `papyros.constants.themes`.

We even provide a theme picker if desired.
image

## Future work
- https://github.com/dodona-edu/papyros/issues/881

closes #849


chvp:
Before I get into the code: 

- I had a weird issue when testing manually where the Run button would not transition back from the "Stop" button into the "Run" button. This only occurred sometimes and required a (very) fast execution of the code.
- When switching to the doctests example, the button for running the doctests only shows up after I do a first execution. If I then switch to another example, the button doesn't disappear again until I run that other example.

jorg-vr:
> I had a weird issue when testing manually where the Run button would not transition back from the "Stop" button into the "Run" button. This only occurred sometimes and required a (very) fast execution of the code.

I haven't 'fixed' the root cause. As I haven't found it yet. But I found the deadlocking loop (Which I had taken over from existing code), and added a relaunch after 5 seconds instead of the infinite loop. So at least if this case occurs, it will be fixed after 5 seconds.

jorg-vr:
> One other question: There seems to be a lot of output printed in the console that wasn't printed on main (all the events that are published?). I get that this can be useful during development, but can we make sure that is disabled for a production build?

This is the existing custom logging, which doesnt run in production

```
const ENVIRONMENT: string = process.env.NODE_ENV || "development";
// Log everything in development
const SHOULD_LOG = ENVIRONMENT !== "production";

/**
 * Helper method to log useful information at runtime
 * @param {LogType} logType The importance of this log message
 * @param {any[]} args The data to log
 */
export function papyrosLog(logType: LogType, ...args: any[]): void {
    const doLog = SHOULD_LOG || logType !== LogType.Debug;
    if (doLog) {
        if (logType === LogType.Error) {
            console.error(...args);
        } else {
            console.log(...args);
        }
    }
}
```

PR #905, opened 2025-10-06
Make sure we don't get infinitely stuck in "running" after input

### LINKED ISSUE #900: Asking for input makes code run forever
Rrunning a code snippet that uses `input` in the new version of Papyros makes the code run forever. Stopping it manually also takes a while.

Image

chvp:
This PR fixes the problem where we could sometimes get stuck in the "Running" state after asking for input. This could happen when the backend finished running before the state was set to running in `provideInput`. By setting it back to running before writing the message we can avoid the ordering issue.

I also tweaked the solution for the similar problem we discovered in #878. Instead of making an exception for when we're going into the running state we now rely on the ordering of the Start and End events. This also means we're only setting back to the ready state on error in the `start` function (because in that case we're not getting a regular "end" event with "CodeFinished" as data).

Fixes #900


TomNaessens:
Managed to reproduce the issue on main, and I was not able to reproduce it anymore on this branch, so manual testing is successful!

On the issue we looked at this afternoon: I got it to work by clicking this box in Chrome, not sure what it does exactly. I guess my service worker was somehow "outdated" (whatever that might mean?).

image

The code changes look good, although I'm not going to pretend I understand everything 😅

TomNaessens:
Did another test this morning (unrelated to input) since termination is handled differently now, and it seems like there's a regression in the handling of exceptions? When writing faulty code, the main branch stops immediately while this branch resumes running.

On the left is production, on the right is this branch:

https://github.com/user-attachments/assets/aab527f8-dcac-42b7-b49f-f5f79cbb1d22

PR #998, opened 2026-04-13
Add Turtle support

TomNaessens:
This PR adds Turtle support to Papyros.

The integrations is based on svg-turtle (https://github.com/donkirkby/svg-turtle) which is what's used in the Turtle judge.

It's 100% seamless: `import turtle` is caught in the python module, and emits svg, drawn by the frontend. All Papyros futures are supported: regular running, debugging (including stepping through code), and `input()`.

https://github.com/user-attachments/assets/d8a08ffd-2075-45e6-9d32-ac6ee93a7aa2

When executing Turtle code, a new tab is added to the output pane and is switched to automatically. In case there's stdout, the user can switch back to output. Once a tab pane has been selected, it's persisted if the code is rerun, or when navigating through debugger steps.

To indicate the size of the canvas, a border is added to the Turtle output.

Since the tabs were floating in space, I added a background colour to the stdout output to connect them to the output.

### DIFF
--- a/src/backend/workers/python/build_package.py
+++ b/src/backend/workers/python/build_package.py
@@ -4,6 +4,7 @@ import shutil
 import os
 import subprocess
 import sys
+import sysconfig
 
 def tarfile_filter(tar_info):
     name = tar_info.name
@@ -27,9 +28,10 @@ def create_package(package_name, dependencies, extra_deps):
     except Exception as e:
         # Always seems to result in a harmless permission denied error
         pass
-    # Bundle CPython's turtle.py (removed from Pyodide's stdlib, required by svg-turtle)
-    import turtle as turtle_mod
-    shutil.copy(turtle_mod.__file__, os.path.join(package_name, "turtle.py"))
+    # Bundle CPython's turtle.py (removed from Pyodide's stdlib, required by svg-turtle).
+    # Locate via sysconfig rather than `import turtle`, which would pull in tkinter.
+    turtle_src = os.path.join(sysconfig.get_path("stdlib"), "turtle.py")
+    shutil.copy(turtle_src, os.path.join(package_name, "turtle.py"))
     tar_name = f"{package_name}.tar.gz.load_by_url"
     if os.path.exists(tar_name):
         os.remove(tar_name)
--- a/src/backend/workers/python/build_package.py
+++ b/src/backend/workers/python/build_package.py
@@ -27,6 +27,9 @@ def create_package(package_name, dependencies, extra_deps):
     except Exception as e:
         # Always seems to result in a harmless permission denied error
         pass
+    # Bundle CPython's turtle.py (removed from Pyodide's stdlib, required by svg-turtle)
+    import turtle as turtle_mod
+    shutil.copy(turtle_mod.__file__, os.path.join(package_name, "turtle.py"))
     tar_name = f"{package_name}.tar.gz.load_by_url"
     if os.path.exists(tar_name):
         os.remove(tar_name)
@@ -45,4 +48,4 @@ def check_tar(tarname, out_dir="."):
 
 
 if __name__ == "__main__":
-    create_package("python_package", "python-runner friendly_traceback pylint>=4,<5 tomli typing-extensions json-tracer>=0.7.0", extra_deps="papyros")
+    create_package("python_package", "python-runner friendly_traceback pylint>=4,<5 tomli typing-extensions json-tracer>=0.7.0 svg-turtle", extra_deps="papyros")
--- a/src/backend/workers/python/papyros/papyros.py
+++ b/src/backend/workers/python/papyros/papyros.py
@@ -103,7 +103,7 @@ class Papyros(python_runner.PyodideRunner):
             if svg_string and svg_string != self._last_emitted_turtle_svg:
                 self._last_emitted_turtle_svg = svg_string
                 img = base64.b64encode(svg_string.encode("utf-8")).decode("utf-8")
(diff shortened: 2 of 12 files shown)


src/communication/BackendEvent.ts
```diff
@@ -14,6 +14,7 @@ export enum BackendEventType {
     FrameChange = "frame-change",
     Stop = "stop",
     Files = "files",
+    Turtle = "turtle",
```
TomNaessens, line 17: note: added this type so we can act on turtle images vs. regular images. Previously in this branch, we did comparison based on the content type (img -> regular image, svg -> turtle images), but I found that too brittle.

src/frontend/state/InputOutput.ts
```diff
@@ -77,6 +82,10 @@ export class InputOutput extends State {
     @stateProperty
     activeEditorTab: string = CODE_TAB;
     @stateProperty
+    activeOutputTab: OutputTab = OUTPUT_TAB;
```
TomNaessens, line 87: note: I thought it would be nice to keep track of the active tab so that when a user is debugging (or rerunning code), we wouldn't switch tabs without their input. Especially useful if you step through programs.

pdawyndt:
WOW! https://en.wikipedia.org/wiki/Wow!_signal

This will definitely be something we can reach out with

- this replaces and the Turtle support from Trinket with additional possibility for debugging (and probably also testing in the future)
- this is part of the GCSE curriculum, so included in many courses (Time2Code as an example)

src/backend/workers/python/papyros/pylint_turtle_brain.py
```diff
@@ -0,0 +1,59 @@
+# Astroid brain plugin for Python's turtle module.
```
TomNaessens, line 1: note: note sure if this is the cleanest thing to teach pylint about the Turtle methods 🤔

src/frontend/components/Output.ts
```diff
@@ -97,11 +136,24 @@ export class Output extends PapyrosElement {
     }
 
     get renderedOutputs(): TemplateResult[] {
-        return this.outputs.map((o) => {
+        let outputsToRender: OutputEntry[];
+        if (this.papyros.io.activeOutputTab === TURTLE_TAB) {
+            // Latest snapshot within this.outputs (which is sliced by the debugger's current
+            // step via maxOutputLength) — so stepping the debugger shows the drawing build up.
+            const lastIdx = this.outputs.findLastIndex((o) => o.type === OutputType.turtle);
+            outputsToRender = lastIdx >= 0 ? [this.outputs[lastIdx]] : [];
```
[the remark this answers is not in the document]
TomNaessens: Right, we'll add a browserlists in another PR. For this one, we can polyfill this in Dodona.

src/backend/workers/python/build_package.py
```diff
@@ -27,6 +27,9 @@ def create_package(package_name, dependencies, extra_deps):
     except Exception as e:
         # Always seems to result in a harmless permission denied error
         pass
+    # Bundle CPython's turtle.py (removed from Pyodide's stdlib, required by svg-turtle)
+    import turtle as turtle_mod
+    shutil.copy(turtle_mod.__file__, os.path.join(package_name, "turtle.py"))
```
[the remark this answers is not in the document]
TomNaessens: Alright, done!

src/backend/workers/python/papyros/papyros.py
```diff
@@ -74,6 +78,7 @@ def cb(typ, dat, contentType=None, **kwargs):
             elif event_type == "input":
                 return cb("input", data["prompt"])
```
chvp, line 80: Should we emit a turtle snapshot in this case as well? We pause the execution until the input is entered, so the user might notice that the picture doesn't match what has been executed so far.

src/frontend/components/Output.ts
```diff
@@ -97,11 +136,24 @@ export class Output extends PapyrosElement {
...
+        } else {
+            outputsToRender = this.outputs.filter((o) => o.type !== OutputType.turtle);
+        }
+        return outputsToRender.map((o) => {
             if (o.type === OutputType.stdout) {
                 return html`${o.content}`;
             } else if (o.type === OutputType.img) {
-                return html`<img src="data:${o.contentType}, ${o.content}"></img>`;
+                const mimeType = o.contentType?.replace(/^img\//, "image/") ?? "image/png";
+                return html`<img src="data:${mimeType},${o.content}"></img>`;
```
chvp: This seems like a band-aid fix for something else? Are we emitting `img/`-prefixed mimetypes somewhere?

TomNaessens: Hmm, yeah, for matplotlib:

```
            def show():
                buf = BytesIO()
                matplotlib.pyplot.savefig(buf, format="png")
                buf.seek(0)
                # encode to a base64 str
                img = base64.b64encode(buf.read()).decode("utf-8")
                matplotlib.pyplot.clf()
                self.output("img", img, contentType="img/png;base64")
```

I don't see why though, I'll patch it here since we're touching mime types anyway for turtle.

TomNaessens: Done in 928efdd6e5ac0d7b4e1c60abde7d58aebc83e216!

src/backend/workers/python/papyros/papyros.py
TomNaessens: Yes! Done in 88a51bf5f182e8792e45ae30e73dcb768bbcd6db

chvp (approved):

PR #999, opened 2026-04-14
Tweak layout and behaviour of tab bar

chvp:
- Font of active tab is bold
- A border was added to the bottom of the tabs, which is different for the active tab
- Blur of the add input field now confirms if text was entered, cancels if nothing was entered
- The edit icon was made clearer
- Tooltips were added to all actions that were just icons
- A confirm step was added when removing a file
- The add input now doesn't pass on the enter press to confirm


TomNaessens (approved):
Looks good to me, but I'll let Bart have the final say 😄

bmesuere (approved):

PR #1003, opened 2026-04-16
Give the editor tab bar a specific separate background

chvp:
What it looks like in papyros:
image
image

What it looks like in Dodona:
image
image


TomNaessens (approved):

bmesuere (changes requested):
Not sure this is the best solution to my remark. Maybe a hairline border on the left/top/right of the tab would also separate it? Or swap the tab colors: make the active tab colored and the others transparent?

bmesuere (approved):
👍

PR #1039, opened 2026-06-02
Fix layout of friendly error in output

### LINKED ISSUE #1004: Cleanup display of error messages
**Problem** This is a screenshot Matthias took from the "feedback" generated by [QuBits](https://qubitsedu.com/). This platform has a cloud IDE for some programming languages and has some learning tracks (probably also allows to create custom learning tracks), but has no support for automated assessment.

Image

The feedback here is nothing but the default Python error message that reports the syntax error, completed with a more accessible explanation of the error message by the FriendlyPython module. Exactly what we also provide in Papyros, but our styling of the error message doesn't help with clear interpretation. 

Image

**Suggested solution**: Improve styling of error messages in Papyros. The Qubits styling makes a clear difference what part of the information comes from Python and what part comes from FriendlyPython, and also uses the fact that FriendlyPython generates Markdown-styled messages.

chvp:
Fixes #1004

Before:
image

After:
image

### DIFF
--- a/src/frontend/components/Output.ts
+++ b/src/frontend/components/Output.ts
@@ -168,18 +168,17 @@ export class Output extends PapyrosElement {
                     const errorObject = o.content as FriendlyError;
                     const errorHTML = [
                         // an array to avoid unintentional spaces/newlines
-                        html`<md-icon title="${errorObject.info}"> ${this.papyros.constants.icons.help} </md-icon>
-                            ${errorObject.name} traceback:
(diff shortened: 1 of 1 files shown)


TomNaessens (approved):

pdawyndt:
Can't we improve this further? The output of FriendlyFeedback is Markdown (see backticks used in the example below), so would be nice to render it at least as Markdown.

Would also be nice to give it a nice visual style as in the example screenshot in #1004

PR #1058, opened 2026-07-25
Allow overriding the backend worker script URL

TomNaessens:
This PR adds `BackendManager.setWorkerUrl(language, url)` plus a thin `papyros.setWorkerUrl(...)` delegate, so a host application can override where a backend's Worker script is loaded from.

For dodona-edu/dodona#8648 the description pages run on the sandbox origin, while assets stay on the main (Claude likes to call this "canonical") host. The initial `new Worker(url)` call must be same-origin (CORS can't relax that), so Dodona currently rewrites its asset host on sandbox requests, which costs cache sharing for every description asset. 

Using this API, Dodona can passes the canonical worker chunk URL instead, so we'll only load the worker chunk once, and not once on the dodona.be ("canonical") url, and then again on the sandbox url. Dodona PR: dodona-edu/dodona#8654.

Before: 

image

After:

image

<details>

<summary>A few notes</summary>

- Papyros owns the mechanism, not just the knob: a same-origin URL is constructed directly, a cross-origin one is bootstrapped through a same-origin blob module that imports the chunk via CORS (the blob inherits the page origin, so the worker stays cookieless on that origin). An earlier draft of this PR exposed a `setWorkerFactory(language, () => Worker)` hook instead, but that put the generic blob trick in the consumer; with the URL API every consumer on a split-origin setup gets it for free, for the JavaScript backend too.
- The blob bootstrap URL is cached per chunk URL (a replaced worker URL simply misses the cache).
- The TSDoc states the full consumer contract: CORS headers on the chunk's origin, and a page CSP whose `worker-src` allows `blob:` plus that origin (Chromium checks a module worker's imports against `worker-src`, not `script-src` — found the hard way, see dodona-edu/dodona#8654).
- Setting a URL drops the cached client for that language (mirroring `removeBackend`, without terminating an already-running worker), so it should be called before launch. Without a registered URL nothing changes.
- The literal `new Worker(new URL(..., import.meta.url), { type: "module" })` expressions stay inlined: host bundlers statically detect exactly that idiom to emit the worker chunk, so it must not be extracted into a variable. There's a comment guarding this, and a small source test (`WorkerIdiom.test.ts`) that fails with the convention spelled out if a refactor breaks it (dev and tests serve modules unbundled, so nothing else would catch it before a consumer's production build).
- The input channel's scope is now anchored to the page origin: sync-message turns it into the base of a synchronous XHR inside the worker, and the relative `"/"` scope is an invalid base URL inside a blob-bootstrapped worker (found by end-to-end testing the split-origin setup in Dodona; `input()` died with `Invalid URL`). Absolute and relative resolve identically for directly-constructed workers.

</details>

Manual testing:
- Run Papyros and assert it just works(tm)
- Load the WIP branch of [#8654](https://github.com/dodona-edu/dodona/pull/8648) and assert that
  - Papyros still loads inside dodona, both in the sandbox, but also in the interactive descriptions
  - Clean out the cache (don't disable it: we want it to use the cache the second load), open an exercise with interactive descriptions, assert it loads the worker chunk on the dodona.be domain, then open up the sandbox and assert it loads the chunk again, cached, from the dodona.localhost host. Compare this to Papyros main where it used to load the first chunk from dodona.localhost, and the second one from sandbox.dodona.localhost.

<details>

<summary>Testing</summary>

- `vitest run test/__tests__/BackendManager.test.ts`: default fallback unchanged, same-origin direct construction, cross-origin blob bootstrap (blob content verified via fetch), cached-client drop on override, bootstrap revocation on URL replacement.
- `test/__tests__/WorkerIdiom.test.ts` guards the bundler convention itself: it asserts the two literal `new Worker(new URL(...))` expressions are present in the source, since a refactor that extracts them passes dev and tests (modules are served unbundled there) and only breaks a consumer's production build with a 404 on the worker chunk.
- `tsc --noEmit` clean; the emitted `dist/communication/BackendManager.js` still contains both literal `new Worker(new URL(...)` expressions verbatim.
</details>

### DIFF
--- a/src/communication/BackendManager.ts
+++ b/src/communication/BackendManager.ts
@@ -24,6 +24,16 @@ export abstract class BackendManager {
      */
     private static backendMap: Map<ProgrammingLanguage, SyncClient<Backend>>;
     /**
+     * Map programming languages to overridden Worker script URLs, consulted by the
+     * default backendCreators registered in the static initializer
+     */
+    private static workerUrls: Map<ProgrammingLanguage, URL> = new Map();
+    /**
+     * Map a Worker script URL's href to a cached same-origin blob bootstrap URL, used when
+     * the URL is cross-origin
+     */
+    private static blobBootstrapUrls: Map<string, string> = new Map();
+    /**
      * Map an event type to interested subscribers
      * Uses an Array to maintain order of subscription
      */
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dodona/papyros",
-  "version": "4.8.0",
+  "version": "4.9.0-rc.2",
```
TomNaessens: TODO before merge: drop this, we'll just merge and release on main

test/__tests__/WorkerIdiom.test.ts
TomNaessens, line 1: note: added this test since the behaviour of linking up the service worker is very specific, and could be easily refactored code-wise, which would lead to it not working no more.

src/frontend/state/Papyros.ts
```diff
@@ -68,7 +74,12 @@ export class Papyros extends State {
             }
             try {
                 await navigator.serviceWorker.register(this.serviceWorkerName, { scope: "/" });
-                BackendManager.channel = makeChannel({ serviceWorker: { scope: "/" } })!;
+                // The channel's scope becomes the base of a synchronous XHR inside the worker;
+                // a relative scope resolves against the worker's own base URL, which is opaque
+                // in a worker bootstrapped from a blob (see BackendManager.setWorkerUrl)
+                BackendManager.channel = makeChannel({
+                    serviceWorker: { scope: `${window.location.origin}/` },
```
bmesuere: instead of the interpolated string, this could also be the `scope` of the service worker that was registered above on line 76

test/__tests__/BackendManager.test.ts
```diff
@@ -35,9 +33,59 @@ describe("BackendManager", () => {
...
+        });
+        vi.stubGlobal("Worker", workerSpy);
+    });
+
+    afterEach(() => {
+        vi.unstubAllGlobals();
+    });
+
+    it("falls back to the default Worker when no URL is set for the language", () => {
+        BackendManager.getBackend(ProgrammingLanguage.Python);
```
bmesuere, line 58: this doesn't actually test the fallback. The toBeCalled also passes when an override is set.

as an aside: I don't think worker urls are reset between tests here?

src/communication/BackendManager.ts
```diff
@@ -124,24 +180,32 @@ export abstract class BackendManager {
...
+                    const url = BackendManager.workerUrls.get(ProgrammingLanguage.Python);
+                    if (url) {
+                        return BackendManager.createWorker(url);
+                    }
+                    // Bundlers detect this exact `new Worker(new URL(...), ...)` expression to emit
+                    // the worker chunk, so the URL and options must stay inlined, not extracted
+                    return new Worker(new URL("../backend/workers/python/worker", import.meta.url), {
+                        type: "module",
+                    });
+                }, BackendManager.channel),
```
bmesuere: identical code is below, so it could be extracted to a workerFactory or something.

bmesuere (approved):

src/frontend/state/Papyros.ts
TomNaessens: Done!

test/__tests__/BackendManager.test.ts
TomNaessens: Also done!

src/communication/BackendManager.ts
TomNaessens: Extracted a bit to a helper method, but had to leave the `new Worker` part in since extracting it broke the vite worker detection otherwise.

bmesuere:
🤖 Heads-up for whoever merges this: this PR conflicts with the #1085 → #1086 stack in `src/frontend/state/Papyros.ts`.

Trial merges against the current tips (both stack branches were just rebased on `main`):

| merge | result |
|---|---|
| #1058 + #1085 | clean |
| #1058 + #1086 | **conflict** in `src/frontend/state/Papyros.ts` |

Everything else merges cleanly — `BackendManager.ts`, `package.json`, `Library.ts` and both test files are touched only by this PR, and the stack's other 12 files are untouched here. Both branches add the same `import { ProgrammingLanguage }` line, which git resolves on its own.

### The conflict

Both sides change the same `configureInput` method:

- **#1058** patches it in place — capture the registration and use `makeChannel({ serviceWorker: { scope: registration.scope } })` instead of `scope: "/"`, because a relative scope resolves against the worker's own base URL, which is opaque in a worker bootstrapped from a blob.
- **#1086** dissolves `configureInput` entirely into `canDeferChannel()` / `ensureChannel()` / `createChannel()`.

So git sees one side editing a method the other side deleted.

### Worth watching when resolving

The natural resolution — "#1086 restructured this, take its version" — **silently drops this PR's scope fix**. #1086's replacement still hardcodes the old form:

```ts
await navigator.serviceWorker.register(this.serviceWorkerName, { scope: "/" });
BackendManager.channel = makeChannel({ serviceWorker: { scope: "/" } })!;
```

Nothing catches that: it compiles, and it only misbehaves in the blob-worker case this PR exists to support, which no test covers (the tests here live in `BackendManager.test.ts` and don't exercise that path).

The two changes also compound — #1058 makes the worker's base URL opaque, #1086 defers registration until a backend asks for one, so the combined code reaches the scope resolution later and from a different call site.

**Resolution:** keep #1086's structure and port the `registration.scope` fix into `createChannel()`, carrying the comment that explains why.

Whichever of the two lands on `main` second will hit this.

package.json
TomNaessens: Done!

PR #1065, opened 2026-07-28
Expose when the backend is ready to run code

### LINKED ISSUE #1064: Expose when the backend is actually ready to run code
`Runner.launch()` deliberately does not await the backend, so a run can be queued while Pyodide is still downloading:

```ts
// src/frontend/state/Runner.ts
public async launch(): Promise<void> {
    this.setState(RunState.Loading);
    const backend = BackendManager.getBackend(this.programmingLanguage);
    // Use a Promise to immediately enable running while downloading
    this.backend = new Promise(async (resolve) => {
        await workerProxy.launch(proxy(...), this.pyodideAssetURL);
        this.updateRunModes();
        return resolve(backend);
    });
    this.setState(RunState.Ready);   // Ready, but the backend is not up yet
}
```

The queueing behaviour is good, but `RunState.Ready` is currently the only signal there is, and it conflates two different things: "you may queue a run" and "the runtime can execute code". Nothing observable distinguishes them.

Two consequences:

1. **UI**: `ButtonLint` renders Run whenever the state is `Ready`, so a user who clicks it during the download drops straight into a bare "Loading" spinner with no explanation of what is being waited for.
2. **Consumers/tests**: there is nothing to wait on. `runner.backend` looks like the signal, but it starts out as a dummy `Promise.resolve({})` and is only swapped for the real promise inside `launch()`, so awaiting the property at an arbitrary moment can resolve immediately and prove nothing.

This bit dodona: its system tests used the Run button appearing as proof the sandbox was usable, and flaked on CI whenever the cold ~12 MB Pyodide download outlasted their wait. Worked around in dodona-edu/dodona#8679 by resolving `runner.backend` in dodona's own code after `await papyros.launch()` and reflecting it as a DOM attribute, which is really Papyros's job.

**Proposal**: add an observable `backendReady` to `Runner`, set to `false` at the start of `launch()` and `true` when the backend promise resolves, and reflect it on a host element (e.g. an attribute on `p-code-runner`) so it is visible from the DOM. That gives the UI something to show a "loading Python" hint with, and gives consumers something to wait on.

bmesuere:
Adds `Runner.backendReady`, so consumers can tell "a run may be queued" apart from "the runtime can execute code". Neither existing signal says the latter: `RunState.Ready` is true from the moment launching starts, and `runner.backend` is a dummy resolved promise until `launch()` swaps it.

It is set false when `launch()` starts and true once the worker is up, and `p-code-runner` reflects it as a `backend-ready` attribute on its host. Queueing a run while the runtime downloads still works exactly as before.

This bit dodona, whose system tests read the Run button as proof the sandbox was usable and flaked when a cold Pyodide download outlasted their wait (dodona-edu/dodona#8679).

**Checklist**
- Tests: the state transition in `Runner.test.ts`, the attribute reflection in a new `components/CodeRunner.test.ts`.
- AI: Claude Code wrote the change and the tests.

Closes #1064

### DIFF
--- a/src/frontend/components/CodeRunner.ts
+++ b/src/frontend/components/CodeRunner.ts
@@ -1,6 +1,6 @@
 import { customElement, state } from "lit/decorators.js";
 import { PapyrosElement } from "./PapyrosElement";
-import { css, CSSResult, html, TemplateResult } from "lit";
+import { css, CSSResult, html, PropertyValues, TemplateResult } from "lit";
 import { createRef, ref, Ref } from "lit/directives/ref.js";
 import { CODE_TAB } from "../state/InputOutput";
 import { arrayBufferToBase64, isTextMimeType } from "../../util/Util";
@@ -65,6 +65,16 @@ export class CodeRunner extends PapyrosElement {
                 background-color: var(--md-sys-color-surface-container);
             }
         `;
+    }
+
+    /**
+     * Reflect backend readiness on the host, so surrounding pages can style or
+     * wait on it. Reading the state here keeps it inside the update cycle the
+     * StateController records, so the host re-renders when it flips.
+     */
+    protected override update(changedProperties: PropertyValues): void {
+        this.toggleAttribute("backend-ready", this.papyros.runner.backendReady);
+        super.update(changedProperties);
     }
 
     protected override firstUpdated(): void {
--- a/src/frontend/state/Runner.ts
+++ b/src/frontend/state/Runner.ts
@@ -62,6 +62,13 @@ export class Runner extends State {
     @stateProperty
     public backend: Promise<SyncClient<Backend>>;
     /**
+     * Whether the backend has finished loading and can execute code.
+     * Runs may be started before this is true: they are queued until the
+     * backend is up, so the run state is Ready while it is still loading.
+     */
+    @stateProperty
+    public backendReady: boolean = false;
+    /**
      * Current state of the program
      */
     @stateProperty
@@ -180,6 +187,7 @@ export class Runner extends State {
      */
     public async launch(): Promise<void> {
         this.setState(RunState.Loading);
+        this.backendReady = false;
         const backend = BackendManager.getBackend(this.programmingLanguage);
         // Use a Promise to immediately enable running while downloading
         // eslint-disable-next-line no-async-promise-executor
@@ -191,6 +199,7 @@ export class Runner extends State {
                 this.pyodideAssetURL,
             );
             this.updateRunModes();
+            this.backendReady = true;
             return resolve(backend);
         });
         this.setState(RunState.Ready);
--- a/src/frontend/state/Runner.ts
+++ b/src/frontend/state/Runner.ts
@@ -69,6 +69,10 @@ export class Runner extends State {
     @stateProperty
     public backendReady: boolean = false;
     /**
+     * Identifies the most recent launch, so a superseded one cannot report ready
+     */
+    private launchId: number = 0;
+    /**
      * Current state of the program
      */
     @stateProperty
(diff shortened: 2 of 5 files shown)


src/frontend/state/Runner.ts
```diff
@@ -191,6 +199,7 @@ export class Runner extends State {
                 this.pyodideAssetURL,
             );
             this.updateRunModes();
+            this.backendReady = true;
```
TomNaessens: Claude sees some issues here, not sure if we want to resolve those: since `launch()` can be called multiple times (when we switch programming language), it's possible that a "slow execution" (launching Python) rapidly following a "fast execution" (launching Javacript) can set the `backendReady` flag to true, while Python actually is still booting up. 

To reproduce (in a console):

```javascript
const app = document.querySelector("p-app"), runner = app.papyros.runner;
const host = app.shadowRoot.querySelector("p-code-runner");
const t0 = performance.now(), at = () => `+${Math.round(performance.now()-t0)}ms`;
const ready = () => `backendReady=${runner.backendReady} attr=${host.hasAttribute("backend-ready")}`;

runner.programmingLanguage = "JavaScript";           // launch A, fast
const a = runner.backend;
a.then(() => console.log(`${at()} [A] JS up   <- ORPHANED launch resolves. ${ready()}`));
runner.programmingLanguage = "Python";               // launch B, slow. A is orphaned
const b = runner.backend;
b.then(() => console.log(`${at()} [B] Python up <- current backend usable. ${ready()}`));
```

<details>

<summary>Claude timings</summary>

The race, reproduced in that live app. I drove the running instance via its own p-app element:

r.programmingLanguage = "JavaScript";  // launch A, fast backend
r.programmingLanguage = "Python";      // launch B, slow backend, same tick

Result:
```
┌─────────┬────────────────────────────────────────────────────────────────────────────────────────────────┐
│    t    │                                             event                                              │
├─────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 0 ms    │ both launches fired, backendReady === false                                                    │
├─────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 31 ms   │ JS backend (the orphaned launch A) resolves                                                    │
├─────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 33 ms   │ backendReady === true, backend-ready attribute present on <p-code-runner>, Python not resolved │
├─────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 1060 ms │ Python backend (the actual current runner.backend) finally resolves                            │
└─────────┴────────────────────────────────────────────────────────────────────────────────────────────────┘
```
</details>

For now, it only seems like this can be triggered by switching very rapidly from Javascript to Python, and I'm not even sure a user can do that within the time Javascript gets loaded, since that's almost instant (31 ms), so not sure if we need to chase this.

If we wanted to, it's proposing to keep track of a `launchId` and to only set `backendReady` to `true` when the current launchId matches the launchId in the promise.

src/frontend/components/CodeRunner.ts
```diff
@@ -67,6 +67,16 @@ export class CodeRunner extends PapyrosElement {
         `;
     }
 
+    /**
+     * Reflect backend readiness on the host, so surrounding pages can style or
+     * wait on it. Reading the state here keeps it inside the update cycle the
+     * StateController records, so the host re-renders when it flips.
+     */
+    protected override update(changedProperties: PropertyValues): void {
+        this.toggleAttribute("backend-ready", this.papyros.runner.backendReady);
```
TomNaessens, line 76: question: should we add this to the README since that already contains the p-code-runner element and the state API?

src/frontend/state/Runner.ts
bmesuere: The multi-programming language is not actively maintained so I think this is the least of our worries related to that?

bmesuere: since it was a small change, i added it anyway

src/frontend/components/CodeRunner.ts
bmesuere: done

chvp (approved):

Issue #1081, opened 2026-08-04
Add JSPI as an input transport for the Python backend, and vendor the sync-message stack


bmesuere:
## Problem

`input()` and `time.sleep()` must block a synchronous Python stack in the worker until the main thread supplies a value. Today this goes through three unmaintained packages: `sync-message` (0.0.12), `comsync` (0.0.9) and `pyodide-worker-runner` (1.4.0). `sync-message` picks one of two mechanisms:

- **Atomics**: `SharedArrayBuffer` + `Atomics.wait`. Requires COOP/COEP. Dodona embeds third-party iframes, so COOP/COEP is permanently off the table and this path never runs there.
- **Service worker**: the worker issues a **synchronous XMLHttpRequest** that a service worker intercepts and holds until the main thread POSTs the input. This is what runs in production on Dodona, for every browser.

The service-worker path is a live reliability liability:

- When the SW is not controlling the page (eviction, browser update, multiple tabs racing on registration, private mode quirks), the student gets a traceback telling them to *"Try closing all this site's tabs, then reopening. If that doesn't work, try using a different browser."* (`python_runner`'s `PyodideRunner.ServiceWorkerError`). Papyros string-matches tracebacks to detect this and report it to Sentry (`src/frontend/state/InputOutput.ts:163-176`).
- Dodona has to serve the SW bundle from a dedicated Rails route (`config/routes.rb:315`) and hard-fails scratchpad startup with a visible alert when registration fails (`coding_scratchpad.ts:66-79`).
- Sentry (dodona org, last 90 days) shows the whole family, ~13 events across ~11 users: `Received status 404 from /__SyncMessageServiceWorkerInput__/write` (DODONA-FRONTEND-FQ, 3 users), `Service worker for reading input was not available` (DODONA-FRONTEND-298), `Error launching papyros after registering service worker` (DODONA-FRONTEND-297), and SW-registration failures (DODONA-FRONTEND-27R, 5 users; DODONA-FRONTEND-2AM). Low absolute counts, but each one is a student who could not run code, and the failure is unrecoverable without closing every tab.
- Sync XHR is deprecated platform-wide; the whole mechanism is a hack living on borrowed time.

[...]

PR #1082, opened 2026-08-04
Vendor the sync-message, comsync and pyodide-worker-runner stack

bmesuere:
This pull request vendors `sync-message`, `comsync` and `pyodide-worker-runner` into `src/sync/`, as phase 1 of #1081.

These three packages implement the synchronous channel that lets Python `input()` and `time.sleep()` block a worker until the main thread answers. All three are unmaintained by the same author, and adding a JSPI transport requires changing their internals, so Papyros now owns the code. No behaviour changes, no public API changes: `src/Library.ts` re-exported none of their types.

Two things could not be copied verbatim, because the upstream code was written without TypeScript strict mode and against a webpack `raw-loader` import, neither of which this repo uses:

- `worker` and `workerProxy` became private fields with getters, so `SyncClient` can keep deleting them on `terminate()` without touching the 9 call sites in `Runner.ts`. `readMessage` reassigned its optional `checkTimeout` parameter, now a properly typed local.
- `pyodide_worker_runner.py` now ships inside `python_package.tar.gz` next to `turtle.py`, instead of being written to `/tmp` at runtime. `build_library.sh` compiles with `tsc`, not vite, so a `?raw` import would have leaked into `dist` unresolved and broken bundling for consumers. This was a phase 2 item in #1081, pulled forward out of necessity.

Also folded in the latent bug #1081 flagged: `BackendManager` built a throwaway channel at import time that `Papyros.configureInput` immediately replaced. Not reachable today, since backends are created lazily after `configureInput`, but it meant any backend constructed earlier would have captured the wrong channel.

**Licensing**

All three are MIT and copyright (c) 2022 Alex Hall, with identical license text. The notice is retained per file, in `THIRD-PARTY-NOTICES.md`, and in the README. Note that `.npmignore` was `/*` plus `!/dist`, so a root notices file would not have shipped in the npm tarball even though `dist` contains the compiled vendored code. It now has an explicit exception, verified with `npm pack --dry-run`.

No user-visible change is expected. To confirm the service worker path still works end to end:

- Run `print(input())` in the scratchpad, enter a value, confirm it echoes
- Run `import time; time.sleep(2); print("done")`, confirm the delay
- Press Stop during `input()` and during an infinite loop, confirm both recover

**Verification**

<details>
<summary>Suite, typecheck, lint and library build</summary>

| Check | Before | After |
|---|---|---|
| `yarn vitest --run` | 7 files / 74 tests | 9 files / 90 tests, green |
| `yarn typecheck` | clean | clean |
| `yarn lint` | clean | clean |
| `yarn build:lib` | n/a | succeeds, no dangling references to the removed packages |

16 new tests. `test/__tests__/sync/channel.test.ts` asserts the literal wire strings `__SyncMessageServiceWorkerInput__` and `__sync-message-v2__` rather than the constants, so a rename cannot silently break a browser still running a service worker registered by an earlier deploy. It covers both delivery orders (write before read, write during a pending read), the 408 timeout and the `/version` handshake. `test/__tests__/sync/SyncClient.test.ts` covers the state machine, including that interrupting a run which is not waiting for input replaces the worker.

</details>

**Checklist**
- Tests: 16 new tests covering the service worker wire protocol, channel selection and the `SyncClient` state machine
- Docs: README gains a `sync` entry in the structure list and a third-party section; removed a line that claimed input handling works "via `sync-message`", which is no longer a dependency
- Announcement: n/a
- AI: Claude Code did the vendoring, the strict-mode and lint conformance, and wrote the tests

Part of #1081

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,3 @@
 /*
 !/dist
+!/THIRD-PARTY-NOTICES.md
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ yarn add @dodona/papyros
 ### Setup input handling
 
 Running interactive programs in the browser requires special handling of synchronous input.
-Papyros supports two approaches (via [`sync-message`](https://github.com/alexmojaki/sync-message)):
+Papyros supports two approaches:
 
 #### COOP/COEP headers
 Add the following HTTP headers to your server responses:
@@ -126,6 +126,8 @@ The codebase organized into clear layers:
 * `frontend`: all browser-side code
     * `state`: state management (e.g. execution state, debugger, input/output)
     * `components`: visualization of that state, as Lit web components
+* `sync`: synchronous communication between the main thread and the workers, so that Python
+  `input()` and `time.sleep()` can block a worker until the main thread answers
 
 ### Components
 
@@ -184,3 +186,11 @@ yarn build:lib
 # Publish to npm
 yarn publish
 ```
+
+## Third-party code
+
+`src/sync/` and `src/backend/workers/python/pyodide_worker_runner.py` are vendored from
+[sync-message](https://github.com/alexmojaki/sync-message),
+[comsync](https://github.com/alexmojaki/comsync) and
+[pyodide-worker-runner](https://github.com/alexmojaki/pyodide-worker-runner),
+all MIT licensed and copyright (c) 2022 Alex Hall. See [THIRD-PARTY-NOTICES.md](THIRD-PARTY-NOTICES.md).
--- a/THIRD-PARTY-NOTICES.md
+++ b/THIRD-PARTY-NOTICES.md
@@ -1,0 +1,36 @@
+# Third-party notices
+
+Papyros includes source code from the projects listed below. Their license terms are reproduced here as required.
+
+## sync-message, comsync, pyodide-worker-runner
+
+- https://github.com/alexmojaki/sync-message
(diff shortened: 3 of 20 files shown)


bmesuere:
🤖 Addressed in #1083 rather than here, since this PR is deliberately a behaviour-identical vendoring and the fixes are behaviour changes.

**`writeMessage` can hang** — real bug, and worse than the comment suggests. Nothing awaits `Runner.provideInput` (`InputOutput.ts:211` calls it without awaiting), so the hang surfaces as the UI stuck in "awaiting input" with no error and no Sentry event. Reproduced with a test that timed out, then fixed: `_reset` now wakes the queued resolver and `writeMessage` re-checks, throwing the same error the already-idle path throws. Note it keys on `_messageIdBase` rather than `state`, because TypeScript narrows `state` past the guard and cannot see that `_reset` ran during the await.

**`callbacks.other!` in `makeRunnerCallback`** — correct that the non-null assertion is unsound, though it matches the upstream behaviour (upstream called `callbacks.other(...)` unguarded, which throws the same way). No fix needed: `makeRunnerCallback` is unreachable from Papyros and is deleted in #1083, since `papyros.py` has its own callback.

**`makeServiceWorkerChannel` trailing slash** — agreed, fixed in #1083 with a test. Papyros always passes `/`, so it is not reachable here, but it is public API and the docstring does not mention the requirement.

TomNaessens (approved):

PR #1085, opened 2026-08-04
Use JSPI stack switching for Python input and sleep where available

bmesuere:
This pull request makes Python `input()` and `time.sleep()` suspend the wasm stack and await a promise, instead of blocking on the sync-message channel, wherever the browser supports WebAssembly stack switching. This is phase 3 of #1081.

Chrome 137+, Firefox 153+ and Safari 27+ take the new path, so in practice every current engine does. The JavaScript backend keeps using the channel and always will: `prompt()` in the JavaScript worker is called from synchronous user code with no wasm on the stack, so nothing can suspend it. Both transports are permanent.

**How the transport is chosen**

`PythonWorker` asks `can_run_sync()` from inside an `async def` entered through a plain PyProxy call, which is exactly how `run_async` is invoked. One question covers browser support, Pyodide build support and the calling convention, and anything unexpected leaves the channel in place. `papyros.py` suspends on the callback result only when it is a promise, so the same Python serves both transports and `python_runner` needed no changes at all.

**Where the routing lives**

Delivery routes inside `SyncClient._writeMessage` rather than at the call sites, so `interrupt()` keeps deciding from state tracked on the main thread. A worker in a busy loop never services Comlink messages, so asking it whether it is waiting deadlocks. This was found the hard way during the investigation, at a 100 second timeout. Backends now report `reading` and `sleeping` through the extras, which keeps that state accurate under JSPI too, and it is why sleep stays interruptible here rather than being a bare `setTimeout` nobody can cancel.

**What this improves**

Python `input()` and `time.sleep()` no longer need a channel at all. That is the whole point: the channel is either `SharedArrayBuffer` plus `Atomics.wait`, which Dodona can never have because it embeds third-party iframes, or a service worker holding a synchronous XHR, which is the mechanism behind the `__SyncMessageServiceWorkerInput__` failures in Sentry and the "close all this site's tabs" dead end that follows them. Nothing here stops registering that service worker yet, that is #1086, but this is what makes it possible.

Interrupting is unchanged, and both transports already handled it well. Stopping a program waiting on `input()` leaves the worker alive either way: the channel path answers the reader with `{interrupted: true}` and `expose.block` throws `InterruptError`, the JSPI path rejects the pending promise, and `python_runner` maps both onto `KeyboardInterrupt` through `InterruptError = KeyboardInterrupt`. Pyodide stays loaded with its packages installed in both cases. The tests now assert that for both transports rather than only for JSPI, since an earlier version of this description claimed the channel path terminated the worker, and it does not.

Interrupting a busy loop is also unchanged: that still needs the interrupt buffer, which still needs a `SharedArrayBuffer`, so the worker is still replaced. JSPI does not help there and this PR does not pretend otherwise.

**Escape hatch**

`runner.allowJspi = false` forces the channel transport. It exists both as a way out if a browser ships broken stack switching, and so CI keeps covering the channel path for Python, which would otherwise go untested the moment Chromium started taking the JSPI branch. That second reason got stronger after Safari 27 shipped JSPI: no mainstream browser reaches the Python channel path by default any more.

- Run `print(input())`, enter a value, confirm it echoes
- Run `import time; time.sleep(2); print("done")`, confirm the delay
- Press Stop while waiting on `input()`. Confirm it stops cleanly with no stderr, and that running `print('alive')` afterwards works without a reload
- Press Stop during `import time; time.sleep(60)`. Confirm it stops immediately rather than after a minute
- Press Stop during `while True: pass`. Confirm it still recovers
- Repeat the input checks in the JavaScript language, which stays on the channel
- Optionally set `papyros.runner.allowJspi = false` before launching and repeat, to exercise the channel path

**Verification**

<details>
<summary>Suite, typecheck, lint and library build</summary>

| Check | Result |
|---|---|
| `yarn vitest --run` | 17 files / 150 tests, green |
| `yarn typecheck` | clean |
| `yarn lint` | clean |
| `yarn build:lib` | succeeds |

11 new tests in `test/__tests__/state/Jspi.test.ts`. The ones that matter most are debug mode, where `JSONTracer` keeps producing correct frames and globals across a suspended `input()`, and the interrupt tests, which assert the backend object is identical before and after so a silent regression to terminate-and-relaunch cannot pass. Interrupting `time.sleep(60)` has its own test that fails if the cancelled timer is ever left to run out. Two further tests drive input and the interrupt path with `allowJspi` off, covering the channel used by browsers without stack switching.

Note that the pre-existing Python tests now exercise the JSPI path by default on Chromium, which is why the forced-channel tests were added rather than relying on them.

</details>

**Checklist**
- Tests: 11 new end-to-end tests covering both transports, debug mode across a suspend, and interrupting input, sleep and a busy loop
- Docs: n/a
- Announcement: n/a, no visible change beyond `input()` and `sleep()` no longer needing the input service worker
- AI: Claude Code investigated the approach, wrote the implementation and the tests

Part of #1081

### DIFF
--- a/src/backend/Backend.ts
+++ b/src/backend/Backend.ts
@@ -49,7 +49,8 @@ export abstract class Backend {
      */
     protected jspi = false;
     /**
-     * The main thread answer this backend is currently suspended on, if any
+     * Settles the promise this backend is suspended on while it waits for the main
+     * thread to answer, if it is waiting at all
      */
     private pending?: { resolve: (value: any) => void; reject: (reason: unknown) => void };
     /**
--- a/src/backend/Backend.ts
+++ b/src/backend/Backend.ts
@@ -1,5 +1,6 @@
 import { BackendEvent, BackendEventType } from "../communication/BackendEvent";
 import { expose, SyncExtras } from "../sync/expose";
+import { InterruptError } from "../sync/errors";
 import { BackendEventQueue } from "../communication/BackendEventQueue";
 
 export interface WorkerDiagnostic {
@@ -43,6 +44,15 @@ export abstract class Backend {
      */
     protected extras: SyncExtras;
     /**
+     * Whether input and sleep suspend the wasm stack (JSPI) instead of blocking on the channel.
+     * Only Pyodide can do this, and only where the browser supports stack switching.
+     */
+    protected jspi = false;
+    /**
+     * The main thread answer this backend is currently suspended on, if any
+     */
+    private pending?: { resolve: (value: any) => void; reject: (reason: unknown) => void };
+    /**
      * Callback to handle events published by this Backend
      */
     protected onEvent: (e: BackendEvent) => any;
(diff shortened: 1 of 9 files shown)


bmesuere:
🤖 Both addressed in 260fac3.

**`reportStatus` doc** — correct, and it was an accuracy problem rather than wording. The contract is asymmetric: `reading` and `sleeping` announce that the backend started waiting, but only sleep reports its own end through `slept`. Input has no counterpart, because the client knows it answered and moves itself back to running in `_writeMessage`. The comment now says that, and also spells out the consequence of not reporting: the client would not know the backend is waiting and `interrupt()` would replace the worker instead of answering it, which is the whole reason the hook exists.

**`Backend.pending` wording** — rephrased.

TomNaessens:
🤖 Posted on Claude's behalf, from a review of this branch at 04232f5. I ran the suite locally (18 files / 152 tests green, `tsc --noEmit` and `eslint src` clean) and wrote some throwaway probes for the paths the tests don't cover. Four things worth a look.

First, one thing I suspected and disproved, so it is *not* a finding: I expected `_writeMessage` under JSPI to block until the worker's Python next suspends, which would make the 5 second deadlock guard in `Runner.stop()` unreachable. Measured it instead, and `writeMessage` resolves in 0ms and `interrupt()` in 1ms even when the program then burns 3 to 4 seconds of pure CPU with no suspension point. Comlink's reply wins the microtask race against the wasm resume. The transport is fine.

### 1. The "what this improves" claim does not match what the channel already does

The body says:

> Before, this path terminated the worker and reloaded the interpreter.

I could not reproduce that. Forcing `allowJspi = false` on this branch, running `x = input('never answered')` and pressing stop while it waits gives:

```
usesPromiseTransport:   false
stateMessage:           Code interrupted after 0.011 s
same SyncClient:        true
same Worker object:     true
stderr present:         false
output after interrupt: ["alive\n"]
```

Same `Worker` object, no stderr, and the interpreter is still usable straight after. The channel path already survives this: `expose.block()` gets `{interrupted: true}` back and throws `InterruptError`, and `PyodideRunner.pyodide_error` maps it onto `KeyboardInterrupt` through `InterruptError = KeyboardInterrupt`, which is exactly what the JSPI rejection does too.

So the interrupt behaviour looks unchanged rather than improved. The real win seems to be that JSPI needs no channel at all, which is what #1086 cashes in. Worth rewording, since the body is the durable record.

### 2. `receiveMessage` and `interruptMessage` return a boolean nobody reads

Both are documented as returning whether the backend was actually waiting (`Backend.ts:124` and `Backend.ts:139`), but `SyncClient._writeMessage` discards it:

```ts
await (message.interrupted ? proxy.interruptMessage() : proxy.receiveMessage(message.message));
```

So an `interruptMessage()` that returns `false`, meaning nothing was pending, still reports back to `interrupt()` as delivered. Either consume the result (for instance, fall back to terminating the worker when it comes back false, which is what the channel path effectively gets from its message ids) or drop the return value and the doc lines, so the contract does not promise something the only caller ignores.

### 3. Interrupting `time.sleep()` under JSPI is untested

The body makes a point of it:

> it is why sleep stays interruptible here rather than being a bare `setTimeout` nobody can cancel

But only `input()` interruption is covered. The whole `settle`/`clearTimeout` branch of `suspendForSleep` (`Backend.ts:173-181`) never runs in the suite: the sleep test lets the timer expire, and both interrupt tests use `input()`. That is exactly the code path the sentence is defending, so it seems worth one test.

### 4. `suspendForInput` and `suspendForSleep` overwrite `this.pending` unguarded

Both assign `this.pending` without checking whether one is already set. If that ever happened the earlier promise would never settle, and the wasm stack would strand with no timeout and nothing logged. I could not find a way to reach it, since the stack is suspended for as long as `pending` is set, so this is defensive rather than a bug. Still, a guard would turn an invisible hang into a loud error, which seems worth it for a state that would be near impossible to diagnose from a bug report.

TomNaessens:
> 🤖 Posted on Claude's behalf ...

Had a look on these, summarised:

1. (nit) Stale(?) PR description, could get a Claude assisted update to have it match the behaviour
2. (nit) I don't see a way for an end user to reach this path and notice it in the UI 
3. (enhancement) Interrupting sleep under JSPI works as expected, but would be nice to have a test for it to cover it
4. (nit) Isn't as bad as Claude makes it seem since the 5 second deadlock guard catches it, and it's also not reachable by a user

TomNaessens (approved):

bmesuere:
🤖 Addressed 1 and 3 in c813cdb. Agreed on 2 and 4, leaving them.

**1. Stale description** — you were right, and I checked before rewriting it: forcing `allowJspi = false` and stopping a waiting `input()` keeps the same worker, produces no stderr, and the interpreter runs the next program straight away. `python_runner`'s `PyodideRunner.InterruptError = KeyboardInterrupt` handles the channel's `InterruptError` and the rejected JSPI promise identically. "What this improves" now says the interrupt behaviour is unchanged and that the actual win is needing no channel for `input()` and `sleep()` at all, which is what #1086 spends. The forced-channel interrupt test asserts the surviving worker as well now, so the description's claim is covered rather than just asserted.

**3. Sleep interrupt untested** — added. It sleeps for 60 seconds and asserts the stop returns in under 5, so the `clearTimeout` in `suspendForSleep`'s settle branch is what makes it pass; a timer left to expire fails it rather than just being slow. Same worker-identity and no-stderr checks as the input case. Needed a `waitForSleeping` helper, since the runner reports `Running` throughout a sleep and only `SyncClient.state` knows.

**2. Unread boolean** — agreed it is unreachable from the UI, and both callers happen at a moment when the state machine already knows something is pending. Leaving the return values as they are.

**4. Unguarded `pending`** — agreed, and the deadlock guard in `Runner.stop()` covers the visible symptom anyway. Leaving it.

Suite after the changes: 17 files / 150 tests green, `yarn typecheck` and `yarn lint` clean.

PR #1113, opened 2026-08-07
Scope the event bus, backends and channel to each Papyros instance

### LINKED ISSUE #1105: Refactor BackendManager to be instance-scoped (per-Papyros event bus and backends)
🤖 *This issue was written by Claude after an investigation session with @bmesuere. It supersedes #881, which it closes as a duplicate.*

## Verdict

The refactor proposed in #881 is worth doing, and two current pain points are both symptoms of exactly the thing that issue targeted: `BackendManager` is the last piece of global state left after #878.

1. **Every browser test costs ≥3 seconds** because the global state forces a fresh `Papyros` + full Pyodide boot per test.
2. **Multiple executable code blocks in one Dodona description** (dodona-edu/dodona#8755) are flatly impossible on the current globals.

The component layer is already instance-ready (`PapyrosElement` takes a `papyros` property that merely *defaults* to the singleton), so the refactor is smaller than it looks.

## What is actually global today

`BackendManager` (`src/communication/BackendManager.ts`) holds four pieces of static state:

1. **`subscriberMap`** — one event bus for the whole page. Every `new Papyros()` constructor registers ~15 listeners (Runner 6, InputOutput 7, Debugger 5+, CodeEditor 1) and there is **no unsubscribe/dispose API**, so instances accumulate forever.
2. **`backendMap`** — one cached `SyncClient` per language, shared by all instances. `SyncClient.call()` throws unless idle, so two instances of the same language can never run concurrently, and they'd share one Pyodide filesystem.
3. **`channel`** — a single input channel, assigned by whichever instance configured input last.
4. **`halted`** — a global "a run ended" flag with special-case exemptions (FrameChange, Files) baked into `publish()`. One instance finishing a run would gate event delivery for every other instance.

The worker→frontend direction is *already* per-instance: `Runner.launchBackend()` passes a Comlink-proxied callback into the worker, which currently just forwards into the static `publish()`. Redirecting that one line to an instance-owned bus is the heart of the refactor.

## Why every test costs ≥3 seconds

- Each test does `new Papyros(); await papyros.launch()` because sharing an instance isn't safe — stale subscribers from previous tests keep reacting to events (e.g. an old `Debugger` can call `papyros.runner.stop()` on frame overflow). The global bus *forces* fresh instances, and `sequence.concurrent: false` + `fileParallelism: false` in `vite.config.ts` exist to keep this pollution sequential.
- `Runner.launch()` fetches the **cached** worker from `backendMap`… and then unconditionally calls `workerProxy.launch()` again. `PythonWorker.launch()` re-runs `loadPyodideAndPackage()` from scratch — a full interpreter re-instantiation inside the same worker. Only the package tarball bytes are cached (`sync/pyodide.ts`). That's the ~3s.
- Worse: the default language is Python, so even JavaScript-only tests (most of `InputOutput.test.ts`) pay a full Pyodide boot during `papyros.launch()` before switching language.

The refactor fixes this via three independent levers:

1. **Idempotent launch** — track "already launched" per worker so a second `launch()` on a live client is a no-op. Cheap, huge win, arguably worth doing standalone.
2. **Per-instance bus + a `dispose()` method** — makes it *safe* to either reuse one Papyros across tests or create fresh ones without leaking subscribers; also unlocks re-enabling concurrent tests and file parallelism.
3. **Injectable backend** — `LaunchFailure.test.ts` already hacks a fake in through `registerBackend`. With a constructor-level injection point, most state-layer tests (InputOutput formatting, Debugger frame handling, Test, Files) could run against `MockBackend` in milliseconds; only Runner/worker integration tests genuinely need Pyodide.

## What multiple code blocks per description (dodona-edu/dodona#8755) need

Multi-instance is impossible today: instance A's output events would be rendered by instance B's InputOutput, and `halted` plus the shared `SyncClient` finish the job. Requirements:

- Per-instance event bus (same change as above).
- Per-instance `SyncClient`/worker. Don't try to multiplex one Pyodide across blocks — the sync layer is built around one blocking call per worker, and blocks would share filesystem state. Instead, **lazy-launch per block on first Run click** (good UX anyway — a description with 5 blocks must not boot 5 Pyodides on page load). #1086's deferred-channel work already moves in this direction.
- One **shared, module-level** service worker registration (a page can only have one per scope anyway) — refcounted or simply idempotent. The channel itself: a service-worker channel can plausibly be shared since message IDs are UUIDs, but an atomics channel wraps a single SharedArrayBuffer and should be per-client. This genuinely needs a test (checkbox 3 of #881).
- `dispose()` also matters here: Dodona descriptions live in Turbo-navigated pages, so blocks get torn down and recreated.

## Scope

**Include:**

- [ ] Move the event bus and backend ownership into `Papyros` (or `Runner`); keep only the *registration config* static — language → worker factory, including `setWorkerUrl` from #1058. Configuration can stay global; state cannot.
- [ ] Idempotent `launch()` per worker; language switch reuses live clients.
- [ ] `dispose()` (unsubscribe + terminate worker).
- [ ] Stop publishing frontend-originated events on the bus (checkbox 2 of #881): `Runner.start()` publishes `Start`/`End` and then *subscribes to its own events* to drive state; replace with direct calls so `halted` and its exemption list can be deleted. `BackendEventType.Stop` has a subscriber but no publisher anywhere in `src` — dead, remove it (after checking Dodona doesn't publish it externally).
- [ ] Keep the `papyros` singleton export in `Library.ts` so Dodona doesn't break; it just becomes "a default instance" rather than a requirement.
- [ ] Test the service worker + channel with multiple instances (checkbox 3 of #881).
- [ ] Re-enable concurrent tests / file parallelism where possible; move state-layer tests to `MockBackend`.

**Exclude:**

- The `BackendEventQueue` worker-side flushing (checkbox 4 of #881): it's orthogonal, it still serves a purpose (batching high-rate output so Comlink isn't hammered per-`print`), and the frontend now has its own batching (Debugger's 50ms flush). Validate it separately — one observation for that later ticket: `shouldFlush()`'s `queue.length > 1` means *any* event-type change flushes immediately, so a debug run interleaving Frame and Output events effectively defeats the batching.
- The `sync/`channel layer (vendored, and #1085 just rewrote `SyncClient` internals).
- Any shared-Pyodide/worker-pool ambitions, and the UI PRs (#902/#903).

## Sequencing

#1085 (JSPI), #1086 (lazy service worker) and #1058 (worker URL override) all modify `Runner`, `Papyros`, `SyncClient` and/or `BackendManager` — rebasing them over this refactor would be miserable, and #1085 even changes interrupt semantics (worker survives input interrupts, so Pyodide stays loaded — which helps the test suite on its own). #1103 and #1098 are Debugger-adjacent and also queued.

**Land those five first, then do the refactor in one focused PR.** The refactor in turn should land *before* any Dodona work on dodona-edu/dodona#8755 starts, since multi-block embedding is unbuildable on the current globals.

bmesuere:
This pull request makes every `new Papyros()` self-contained. `BackendManager` was the last global after #878: one subscriber map, one client cache, one channel and one `halted` flag for the whole page. Each instance now owns its event bus (`papyros.events`), its backend clients and its channel, which is what dodona-edu/dodona#8755 (multiple executable code blocks per description) needs, and what lets tests inject or reuse backends safely. Rebased onto main now that #1085 and #1086 are in.

**What moved where**

- `EventBus` is a new instance-scoped pub/sub whose `subscribe` returns an unsubscribe function. Nothing accumulates across instances anymore, and components (`p-code-editor`) unsubscribe on disconnect.
- `BackendManager` keeps only the static registration config (language to worker factory). Clients are created per instance by `Runner`, and `runner.registerBackend` overrides the factory per instance, so `LaunchFailure.test.ts` no longer mutates global state to inject its fake.
- The channel is `papyros.channel`. The one service worker registration a page can have is shared between instances through a module-level promise; each instance builds its own channel on top of it.
- Frontend-originated events no longer ride the bus. `start()` and `stop()` call `io.reset()`, `io.onRunEnd()`, `debugger.onRunStart()` and `debugger.onRunEnd()` directly, so the `halted` gate and its FrameChange/Files exemptions are deleted. The `Stop` and `FrameChange` event types had no publisher anywhere (and `BackendManager` was never part of the public API, so Dodona cannot be publishing them) and are removed.
- `launch()` is idempotent per worker: relaunching a live client is a cache hit, and language switches reuse live clients. The launch cache is keyed on the `Worker` itself, so an interrupt that replaces the worker invalidates it by construction.
- `dispose()` terminates an instance's workers, for Turbo-navigated pages that tear blocks down and recreate them.

**Deliberate behavior change**

Output the worker had already buffered when the user pressed stop is now delivered instead of dropped, since the `halted` gate that suppressed it is gone. The interrupt itself still surfaces as the usual "Code interrupted" state, not as an error.

- Run a Python program with `input()`, answer it, stop a `while True: pass` loop, and switch to JavaScript and back: everything behaves as before.
- The `papyros` singleton export is unchanged, it is simply a default instance now.

**Verification**

| Check | Result |
|---|---|
| `yarn vitest --run` | 20 files / 161 tests, green |
| `yarn typecheck` | clean |
| `yarn lint` | clean |

**Checklist**
- Tests: new `EventBus.test.ts` (delivery, dedupe, unsubscribe, instance isolation) and `MultiInstance.test.ts` (two instances run the same language concurrently with isolated output, per-instance channels over the shared service worker registration, cross-instance event isolation, dispose). Channel/LaunchFailure/Jspi/BackendManager tests updated for instance scoping.
- Docs: README gains a "Multiple instances" section and documents `papyros.events`.
- Announcement: n/a
- AI: Claude Code wrote the implementation and the tests

Closes #1105

### DIFF
--- a/src/frontend/state/Runner.ts
+++ b/src/frontend/state/Runner.ts
@@ -169,17 +169,68 @@ export class Runner extends State {
      */
     private papyros: Papyros;
 
+    /**
+     * The live backend client per language, owned by this instance
+     */
+    private clients: Map<ProgrammingLanguage, SyncClient<Backend>> = new Map();
+    /**
+     * Per-instance factory overrides, so tests can inject a backend without
+     * touching the static registry shared by every instance
+     */
+    private backendCreators: Map<ProgrammingLanguage, () => SyncClient<Backend>> = new Map();
+    /**
+     * Tracks which workers have completed their launch call, so relaunching a live
+     * client is free. Keyed by the Worker itself: an interrupt that replaces the
+     * worker automatically invalidates the entry.
+     */
+    private launched: WeakMap<object, Promise<void>> = new WeakMap();
+
     constructor(papyros: Papyros) {
         super();
         this.papyros = papyros;
         this.backend = Promise.resolve({} as SyncClient<Backend>);
 
-        BackendManager.subscribe(BackendEventType.Input, () => this.setState(RunState.AwaitingInput));
-        BackendManager.subscribe(BackendEventType.Loading, (e) => this.onLoad(e));
-        BackendManager.subscribe(BackendEventType.Start, (e) => this.onStart(e));
-        BackendManager.subscribe(BackendEventType.End, (e) => this.onEnd(e));
-        BackendManager.subscribe(BackendEventType.Error, () => this.onError());
-        BackendManager.subscribe(BackendEventType.Stop, () => this.stop());
+        this.papyros.events.subscribe(BackendEventType.Input, () => this.setState(RunState.AwaitingInput));
+        this.papyros.events.subscribe(BackendEventType.Loading, (e) => this.onLoad(e));
+        this.papyros.events.subscribe(BackendEventType.Start, (e) => this.onStart(e));
+        this.papyros.events.subscribe(BackendEventType.End, (e) => this.onEnd(e));
+        this.papyros.events.subscribe(BackendEventType.Error, () => this.onError());
+    }
+
+    /**
+     * Use a custom backend for the given language on this instance only
+     * @param {ProgrammingLanguage} language The language to override
+     * @param {Function} backendCreator The constructor for a SyncClient
+     */
+    public registerBackend(language: ProgrammingLanguage, backendCreator: () => SyncClient<Backend>): void {
+        this.backendCreators.set(language, backendCreator);
+        this.clients.delete(language);
+    }
+
+    private getClient(language: ProgrammingLanguage): SyncClient<Backend> {
+        let client = this.clients.get(language);
+        if (!client) {
+            const create = this.backendCreators.get(language);
+            client = create ? create() : BackendManager.createBackend(language);
+            this.clients.set(language, client);
+        }
+        return client;
+    }
+
+    /**
+     * Terminate every worker this instance started and abandon in flight launches
+     */
+    public dispose(): void {
+        this.launchId++;
+        this.backendReady = false;
+        for (const client of this.clients.values()) {
+            try {
+                client.terminate();
+            } catch {
+                // An injected or never-started client has no worker to terminate
+            }
+        }
+        this.clients.clear();
     }
 
     /**
@@ -202,7 +253,7 @@ export class Runner extends State {
         this.setState(RunState.Loading);
         this.backendReady = false;
         const launchId = ++this.launchId;
-        const backend = BackendManager.getBackend(this.programmingLanguage);
+        const backend = this.getClient(this.programmingLanguage);
         // Expose the promise before it settles so runs can already be queued while downloading
         const backendLaunched = this.launchBackend(backend, launchId);
         this.backend = backendLaunched;
@@ -218,13 +269,29 @@ export class Runner extends State {
     }
 
     private async launchBackend(backend: SyncClient<Backend>, launchId: number): Promise<SyncClient<Backend>> {
-        // Allow passing messages between worker and main thread
-        await backend.workerProxy.launch(
-            proxy((e: BackendEvent) => BackendManager.publish(e)),
-            this.pyodideAssetURL,
-            this.allowJspi,
-        );
-        backend.usesPromiseTransport = await backend.workerProxy.usesJspi();
+        // An injected test double has no worker, so fall back to keying on the client
+        const worker: object = backend.worker ?? backend;
+        let launched = this.launched.get(worker);
+        if (!launched) {
+            // Allow passing messages between worker and main thread
+            launched = backend.workerProxy
+                .launch(
+                    proxy((e: BackendEvent) => this.papyros.events.publish(e)),
+                    this.pyodideAssetURL,
+                    this.allowJspi,
+                )
+                .then(async () => {
+                    backend.usesPromiseTransport = await backend.workerProxy.usesJspi();
+                });
+            this.launched.set(worker, launched);
+        }
+        try {
+            await launched;
+        } catch (error) {
+            // Let a retry attempt the launch again instead of replaying this failure
+            this.launched.delete(worker);
+            throw error;
+        }
         if (!backend.usesPromiseTransport) {
             // This backend blocks on the channel, so it needs one to exist before it runs.
             // Registration may not have happened yet: Papyros defers it when the browser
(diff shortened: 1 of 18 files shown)


src/frontend/state/Runner.ts
```diff
@@ -169,17 +169,68 @@ export class Runner extends State {
...
+            this.clients.set(language, client);
+        }
+        return client;
+    }
+
+    /**
+     * Terminate every worker this instance started and abandon in flight launches
+     */
+    public dispose(): void {
+        this.launchId++;
```
bmesuere, line 231: 🤖 [P1] Mark the runner as disposed here and use that state to suppress later relaunch/retry paths. If Turbo tears down a block while it is running, `dispose()` terminates the worker; that rejects the active `call()` with `InterruptError`, sets `terminated = true`, and the `finally` block immediately calls `launch()` to create a replacement worker after this method clears the client map. I reproduced it with a JavaScript `while (true) {}` run: after `dispose()` and awaiting the run promise, `backendReady` is `true` again. This leaks exactly the worker disposal is meant to release; please also cover disposal during an active run in the test.

bmesuere (commented):
🤖 Review finding

src/frontend/state/Runner.ts
bmesuere: 🤖 Fixed in 41030e1e: `Runner` now tracks a `disposed` flag and `launch()` refuses to run after disposal, so the interrupted-run path in `start()` can no longer boot a replacement worker. Added the requested regression test: a JavaScript `while (true) {}` run is disposed mid-run, and `backendReady` stays false after the run promise settles.

bmesuere:
🤖 Two commits landed since the review started:

- `41030e1e` fixes the dispose-during-run relaunch the review bot found (see the resolved thread), with a regression test.
- `de6f8386` fixes a bug that #1114's shared-instance tests exposed: the loading event `provideFiles` hands to `onLoad` was a stringified payload, but `parseData` returns `application/json` data as-is, so the status read as undefined, took the failed branch, and its state restore could leave the runner stuck on Loading after `provideFiles`. The `halted` gate used to drop that event between runs, which is why it never surfaced before this PR. The event now carries the object itself.

The test-suite speedup is stacked on top as #1114.

TomNaessens (approved):

PR #1123, opened 2026-08-28
Recover from an exhausted Python runtime

bmesuere:
This pull request lets the scratchpad recover when its Python runtime can no longer run or lint code, instead of silently losing linting for the rest of the session.

Reported as Sentry DODONA-FRONTEND-2CV: a `MemoryError` raised by `NamedTemporaryFile` in `linting.py`, which is the *first* allocation `lint()` makes, so the WebAssembly heap was already full before linting started. `Runner.lintSource()` had no error handling, so `@codemirror/lint` turned the rejected linter into an uncaught `window.onerror`, and nothing ever replaced the exhausted runtime.

Three changes:

- `SyncClient.restart()` replaces the worker. `interrupt()` already did exactly this when no interrupt buffer is available, but it returns early when the client is idle, which it is after a failed lint. `Runner.launch()` on its own does not recycle Pyodide either: it hands back the cached client and reuses its `launched` entry.
- A failed lint now goes to `papyros.errorHandler` and shows no diagnostics, rather than escaping to the page.
- An error that leaves the runtime unusable triggers a restart, a relaunch, and a rewrite of the workspace files, which a fresh worker comes up without while the editor still lists them. Files handed over through `provideFiles` are replayed as well, since `io.files` skips anything over 1 MB. A run suspended in `input()` at that moment has its prompt closed, as `stop()` does. The run path gets the same wiring as a precaution: both Sentry issues in this family so far (DODONA-FRONTEND-2CV and DODONA-FRONTEND-29S) failed on the lint path, and the Safari 18.3 `Out of bounds memory access` events (DODONA-FRONTEND-2BQ) fail inside `loadPyodide`, which goes through the launch path and is unchanged here.
- The signatures that count as an unusable runtime cover V8, JavaScriptCore and SpiderMonkey trap wording (checked against Chromium 151, the macOS `jsc` shell and Firefox 154) and Pyodide's own "already fatally failed" post-mortem.

**Not fixed here: what fills the heap.** Linting leaks roughly 0.5 MB per lint, and the editor lints about 750 ms after every edit, so a long session walks into the ceiling. Fixing that needs private pylint, astroid and isort APIs, which we decided against; students still reach the ceiling, they now recover from it.

<details>
<summary>Measurements behind the leak claim</summary>

Measured natively against the pinned versions (Python 3.14, pylint 4.0.6, astroid 4.0.4), driving this repo's own `linting.py` with its plugins and rc file. 300 lints of an unchanged 20 line program:

| | RSS after 300 lints | live `MessageDefinition` |
|---|---|---|
| as is | 73 to 235 MB, linear, no plateau | 120,000 |
| releasing the caches | 73 to 95 MB, flat from lint 100 on | 400 |

Three caches hold the per lint object graph:

- `MessageDefinitionStore.get_message_definitions` is a `@functools.cache` on a *method*, keyed `(self, msgid)`, so every store ever built stays alive with about 390 `MessageDefinition` objects. That line even carries `# pylint: disable=method-cache-max-size-none`, arguing the cache "will likely stay minimal", which holds for a one shot CLI and not for a worker that lives for a whole session.
- `isort.place.module_with_reason` is an `lru_cache` keyed on `(name, Config)`, pinning the fresh `isort.Config` that pylint's `ImportsChecker` builds per run.
- astroid's `astroid_cache` keeps the linted module, since `lint()` gives every lint a new `NamedTemporaryFile` and so a new module name.

Pylint's own supported release (`--clear-cache-post-run`) is not usable here: it is 14 times slower (290 ms versus 20 ms per lint), its RSS is worse, and it wipes `AstroidManager.brain["_transform"]`, which permanently drops our turtle brain. Verified: after one `clear_cache()`, `turtle.forward(100)` starts reporting "Module 'turtle' has no 'forward' member".

</details>

- Open the scratchpad, run code that exhausts memory (a loop appending to a list until it dies), then keep typing. Linting keeps working and no uncaught error reaches the console.
- Do the same with a file added through the dropzone. After the restart the file is still listed and still readable from Python.

**Checklist**
- Tests: `test/__tests__/state/RuntimeRecovery.test.ts`, six cases. A failed lint resolves instead of throwing and leaves the runtime alone; a `MemoryError` replaces the runtime and restores its files, whether it came from a lint or from a run; a recovery closes the input prompt of the run it abandons; provided files are handed over again; a failed launch stops linting without retrying it. All fail before this change. `SyncClient.test.ts` gains one case for restarting an already terminated client, which fails before this change too.
- Docs: n/a
- Announcement: 🐛 The coding scratchpad now recovers by itself when it runs out of memory, instead of quietly losing its error highlighting.
- AI: Claude Code traced the Sentry issue to its root cause and wrote this change, via the sentry-triage skill.

Reaches production once Dodona bumps `@dodona/papyros`.

### DIFF
--- a/src/frontend/state/Runner.ts
+++ b/src/frontend/state/Runner.ts
@@ -36,17 +36,25 @@ export interface LoadingData {
 
 /**
  * Signatures of errors that leave the runtime unable to run or lint anything
- * afterwards: its heap is exhausted, or a trap aborted it. The WebAssembly heap
+ * afterwards: its heap is exhausted, a trap aborted it, or Pyodide already
+ * declared it dead after a trap that surfaced elsewhere. The WebAssembly heap
  * only ever grows, so nothing short of a fresh worker recovers from these.
+ * Trap wording differs per engine (V8, JavaScriptCore, SpiderMonkey).
  */
 const UNUSABLE_RUNTIME_SIGNATURES = [
     "memoryerror",
     "out of bounds memory access",
     "memory access out of bounds",
+    "index out of bounds",
     "could not allocate memory",
     "call_indirect to a null table entry",
+    "call_indirect to a signature that does not match",
     "null function",
+    "indirect call to null",
+    "indirect call signature mismatch",
     "aborted(",
+    "pyodide already fatally failed",
+    "pyodide already exited",
 ];
 
 function isRuntimeUnusable(error: any): boolean {
@@ -169,7 +177,15 @@ export class Runner extends State {
      * Async getter for the linting diagnostics of the current code
      */
     public async lintSource(): Promise<WorkerDiagnostic[]> {
-        const backend = await this.backend;
+        let backend: SyncClient<Backend>;
+        try {
+            backend = await this.backend;
+        } catch {
+            // Launch failures surface through launch(), and a lint cannot recover
+            // from one: retrying the launch on every edit would loop on a runtime
+            // that fails to boot
+            return [];
+        }
         const proxy = backend.workerProxy;
 
         if (!proxy) {
@@ -225,6 +241,12 @@ export class Runner extends State {
      * old one is still exhausted fails too, and each failure asks for a recovery.
      */
     private recovering: boolean = false;
+    /**
+     * The files last handed over through provideFiles, replayed into a replacement
+     * runtime. Files over 1 MB never reach io.files, so they cannot be restored
+     * from there.
+     */
+    private providedFiles?: [Record<string, string>, Record<string, string>];
 
     constructor(papyros: Papyros) {
         super();
@@ -327,6 +349,10 @@ export class Runner extends State {
         }
         this.recovering = true;
         try {
+            // A run suspended in input() dies with its worker, so close its prompt
+            // and flush its frames the way stop() does
+            this.papyros.io.onRunEnd();
+            this.papyros.debugger.onRunEnd();
             const client = this.clients.get(this.programmingLanguage);
             try {
                 client?.restart();
@@ -347,7 +373,12 @@ export class Runner extends State {
      * empty filesystem while the editor still shows them.
      */
     private async restoreWorkspace(): Promise<void> {
+        // Snapshot first: replaying the provided files refreshes io.files from the
+        // worker, and the editor's copy of a file edited since must win
         const files = this.papyros.io.files;
+        if (this.providedFiles) {
+            await this.provideFiles(...this.providedFiles);
+        }
         if (files.length === 0) {
             return;
         }
@@ -562,6 +593,7 @@ export class Runner extends State {
         if (fileNames.length === 0) {
             return;
         }
+        this.providedFiles = [inlinedFiles, hrefFiles];
         // parseData hands application/json data over as-is, so it must be the object
         this.onLoad({
             type: BackendEventType.Loading,
--- a/src/frontend/state/Runner.ts
+++ b/src/frontend/state/Runner.ts
@@ -35,6 +35,26 @@ export interface LoadingData {
 }
 
 /**
+ * Signatures of errors that leave the runtime unable to run or lint anything
+ * afterwards: its heap is exhausted, or a trap aborted it. The WebAssembly heap
+ * only ever grows, so nothing short of a fresh worker recovers from these.
+ */
+const UNUSABLE_RUNTIME_SIGNATURES = [
+    "memoryerror",
+    "out of bounds memory access",
+    "memory access out of bounds",
+    "could not allocate memory",
+    "call_indirect to a null table entry",
+    "null function",
+    "aborted(",
+];
+
+function isRuntimeUnusable(error: any): boolean {
+    const description = `${error?.type ?? ""} ${error?.message ?? error ?? ""}`.toLowerCase();
+    return UNUSABLE_RUNTIME_SIGNATURES.some((signature) => description.includes(signature));
+}
+
+/**
  * Helper component to manage and visualize the current RunState
  */
 export class Runner extends State {
@@ -155,7 +175,18 @@ export class Runner extends State {
         if (!proxy) {
             return [];
         }
-        return await proxy.lintCode(this.code);
+        try {
+            return await proxy.lintCode(this.code);
+        } catch (error: any) {
+            // The editor lints in the background on every edit, and CodeMirror turns a
+            // rejected linter into an uncaught window error. Report it and show no
+            // diagnostics instead.
+            this.papyros.errorHandler(error);
+            if (isRuntimeUnusable(error)) {
+                await this.recoverRuntime();
+            }
+            return [];
+        }
     }
 
     /**
@@ -189,6 +220,11 @@ export class Runner extends State {
      * interrupted, which normally relaunches the worker; this suppresses it.
      */
     private disposed: boolean = false;
+    /**
+     * Whether a runtime is already being replaced. Every lint that lands while the
+     * old one is still exhausted fails too, and each failure asks for a recovery.
+     */
+    private recovering: boolean = false;
 
     constructor(papyros: Papyros) {
         super();
@@ -281,6 +317,46 @@ export class Runner extends State {
         }
     }
 
+    /**
+     * Replace a runtime that can no longer run or lint code, and put the files it
+     * held back into the fresh one.
+     */
+    private async recoverRuntime(): Promise<void> {
+        if (this.disposed || this.recovering) {
+            return;
+        }
+        this.recovering = true;
+        try {
+            const client = this.clients.get(this.programmingLanguage);
+            try {
+                client?.restart();
+            } catch {
+                // An injected or never-started client has no worker to replace
+            }
+            await this.launch();
+            await this.restoreWorkspace();
+        } catch (error: any) {
+            this.papyros.errorHandler(error);
+        } finally {
+            this.recovering = false;
+        }
+    }
+
+    /**
+     * Write the files back into a freshly started worker, which comes up with an
+     * empty filesystem while the editor still shows them.
+     */
+    private async restoreWorkspace(): Promise<void> {
+        const files = this.papyros.io.files;
+        if (files.length === 0) {
+            return;
+        }
+        const backend = await this.backend;
+        for (const file of files) {
+            await backend.workerProxy.updateFile(file.name, file.content, file.binary);
+        }
+    }
+
     private async launchBackend(
         language: ProgrammingLanguage,
         backend: SyncClient<Backend>,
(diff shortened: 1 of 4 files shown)


src/sync/SyncClient.ts
```diff
@@ -67,6 +67,14 @@ export class SyncClient<T = any> {
...
+    }
+
+    /**
+     * Replace the worker with a fresh one, dropping everything the old one held.
+     * This client keeps its identity, so callers holding it stay valid.
+     */
+    public restart(): void {
         this.terminate();
         this._start();
     }
```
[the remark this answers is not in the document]
bmesuere: Fixed in 2f05645f. `terminate()` now releases the proxy and the worker with optional chaining, so calling it twice is a no-op and `restart()` always leaves a fresh worker behind.

Not reachable through the current call paths (a client that gets terminated is dropped from `Runner.clients` first, and `recoverRuntime()` returns early once disposed), but it is a real sharp edge on a newly public method: `Runner` already wraps both of its `terminate()` calls in try/catch for exactly this reason. Added a regression test that fails with the old code on `TypeError: Cannot read properties of undefined (reading 'Symbol(Comlink.releaseProxy)')`.

src/frontend/state/Runner.ts
```diff
@@ -281,6 +317,46 @@ export class Runner extends State {
...
+     */
+    private async recoverRuntime(): Promise<void> {
+        if (this.disposed || this.recovering) {
+            return;
+        }
+        this.recovering = true;
+        try {
+            const client = this.clients.get(this.programmingLanguage);
+            try {
+                client?.restart();
```
TomNaessens, line 357: 🤖 **Bug:** a recovery triggered by a lint while a run is suspended in `input()` leaves the input prompt open.

With the JSPI transport the worker still services Comlink calls while the program is suspended, so `lintCode` can run and fail mid-run. `restart()` then calls `terminate()`, which rejects the running `call()` with an InterruptError, and `start()` takes its `interrupted`/`terminated` branch. That branch assumes `stop()` already ran `io.onRunEnd()` and `debugger.onRunEnd()`, so neither runs here: the state becomes Ready with the "interrupted" message while `awaitingInput` stays true. Typing into the prompt then calls `writeMessage`, which throws "No active call to send a message to." as an unhandled rejection.

Both `onRunEnd()` methods are idempotent, so calling them here before the restart, as `stop()` does, is enough:

```ts
this.papyros.io.onRunEnd();
this.papyros.debugger.onRunEnd();
client?.restart();
```

src/frontend/state/Runner.ts
```diff
@@ -373,13 +450,16 @@ export class Runner extends State {
...
+                unusable = isRuntimeUnusable(error);
             }
         } finally {
             if (this.state === RunState.Stopping) {
                 // stop() already closed the input prompt and flushed the debugger
                 interrupted = true;
             }
-            if (terminated) {
+            if (unusable) {
+                await this.recoverRuntime();
```
TomNaessens, line 492: 🤖 The Sentry evidence cited for the run path does not match. DODONA-FRONTEND-2BQ (23 events, all Safari 18.3) is thrown from `loadPyodide` → `finalizeBootstrap` → `rawRun("import _pyodide_core")`: Pyodide fails while booting, so the error goes through the `launchBackend` catch (the #1116 path) and never reaches `start()`. This PR changes nothing for it.

The two issues in this family that did occur are both on the lint path: DODONA-FRONTEND-2CV (the `MemoryError` from `NamedTemporaryFile`) and DODONA-FRONTEND-29S (June, `Error in sys.excepthook ... MemoryError` raised from `install_imports` inside `lintCode`). The substring match catches both.

Keeping the run path wiring is harmless, but the description should present it as precautionary rather than cite 2BQ. If it stays, `start()` with an unusable error is the only branch without a test.

TomNaessens (commented):
🤖 Review by Claude Code, posted at Tom's request. Five inline comments below. The first is a bug in the new code path; the other four are lower priority.

src/frontend/state/Runner.ts
```diff
@@ -34,6 +34,26 @@ export interface LoadingData {
     status: "loading" | "loaded" | "failed";
 }
 
+/**
+ * Signatures of errors that leave the runtime unable to run or lint anything
+ * afterwards: its heap is exhausted, or a trap aborted it. The WebAssembly heap
+ * only ever grows, so nothing short of a fresh worker recovers from these.
+ */
+const UNUSABLE_RUNTIME_SIGNATURES = [
```
TomNaessens, line 44: 🤖 Two groups of messages are missing from this list. Neither has appeared in Sentry yet.

- Pyodide's own post-mortem. After a trap, `API.fatal_error` replaces the public API with functions that throw `Pyodide already fatally failed and can no longer be used.` (verified in `pyodide.asm.mjs`, 314.0.3). If the trap itself is raised somewhere this code does not catch (an event callback, a timer), the next lint or run is the first error that reaches `isRuntimeUnusable`, and that message matches nothing in the list. The runtime then stays dead, and every lint reports to Sentry.
- Firefox wording. The list covers V8 and JavaScriptCore. SpiderMonkey reports the same traps as `index out of bounds` (memory), `indirect call to null` / `indirect call signature mismatch` (call_indirect) and `unreachable executed`. These strings are from memory, not verified against a current Firefox, so check them before adding.

src/frontend/state/Runner.ts
```diff
@@ -281,6 +317,46 @@ export class Runner extends State {
...
+            this.recovering = false;
+        }
+    }
+
+    /**
+     * Write the files back into a freshly started worker, which comes up with an
+     * empty filesystem while the editor still shows them.
+     */
+    private async restoreWorkspace(): Promise<void> {
+        const files = this.papyros.io.files;
```
TomNaessens, line 377: 🤖 Provided files larger than 1 MB are not restored. `io.files` is filled from the worker's `files` event, and `_emit_created_files` skips any file over 1 MB, so a dataset of that size handed over through `provideFiles` never appears in `io.files`. After a recovery the fresh worker does not have it: the student's `open("data.csv")` raises FileNotFoundError until the page is reloaded, and nothing explains why. The Runner does not keep the `provideFiles` arguments either, so it cannot replay them.

One fix: store the last `(inlinedFiles, hrefFiles)` on the Runner in `provideFiles`, and replay them here before the `updateFile` loop so a file the student edited afterwards still wins. Files added through the dropzone are unaffected, since `io.upsertFile` stores their content regardless of size.

src/frontend/state/Runner.ts
```diff
@@ -281,6 +317,46 @@ export class Runner extends State {
...
+            const client = this.clients.get(this.programmingLanguage);
+            try {
+                client?.restart();
+            } catch {
+                // An injected or never-started client has no worker to replace
+            }
+            await this.launch();
+            await this.restoreWorkspace();
+        } catch (error: any) {
+            this.papyros.errorHandler(error);
```
TomNaessens, line 364: 🤖 If the relaunch itself fails, the runtime is dead again with no further recovery. `launch()` assigns `this.backend = backendLaunched` before awaiting it (pre-existing), so a failed relaunch leaves `this.backend` holding a rejected promise. `lintSource()` awaits that promise before its try block, so every later lint rejects outside the catch and escapes through CodeMirror's `logException` as an uncaught error, which is the behaviour the new catch in `lintSource()` was added to stop. The initial launch has `Papyros.launch()`'s confirm-and-retry around it; this path has nothing.

A relaunch failing should be rare. At minimum, moving `await this.backend` inside the try in `lintSource()` routes these failures to `errorHandler` like any other lint failure.

src/frontend/state/Runner.ts
bmesuere: 🤖 Fixed in 587cdb25. `recoverRuntime()` now calls `io.onRunEnd()` and `debugger.onRunEnd()` before `restart()`, as `stop()` does. Test: "closes the input prompt of a run the recovery abandons" sets `awaitingInput` and fails a lint with a `MemoryError`; it fails before this change.

src/frontend/state/Runner.ts
bmesuere: 🤖 Description reworded in place: the run path is now presented as precautionary, with 2CV and 29S named as the two lint-path occurrences and 2BQ identified as a `loadPyodide` failure on the launch path. The run path branch is covered by a new test in 587cdb25 ("replaces a runtime that a run left unusable"), which fails a `runCode` call with a `MemoryError` and checks for the restart and the file restore.

src/frontend/state/Runner.ts
bmesuere: 🤖 Both added in 587cdb25.

- Pyodide: `pyodide already fatally failed` and `pyodide already exited`, the two variants of the `Pyodide already ${exited|fatally failed} and can no longer be used.` template in pyodide.asm.mjs 314.0.3.
- SpiderMonkey: `index out of bounds`, `indirect call to null` and `indirect call signature mismatch`. Verified against the `JSMSG_WASM_*` strings compiled into the installed Firefox 154.0.1 (`XUL` binary); `unreachable executed` is there too but left out, since the other engines' unreachable wording is not in the list either and a bare `unreachable` would match a student's own "host unreachable" message.

While checking the other engines with tiny trap modules: the current macOS `jsc` reports a call through a null table entry as `call_indirect to a signature that does not match`, not `call_indirect to a null table entry`, so that wording is added as well. Chromium 151 gives `null function` / `function signature mismatch` / `memory access out of bounds`, which the list already covered.

src/frontend/state/Runner.ts
bmesuere: 🤖 Fixed in 587cdb25 along the lines you suggest. `provideFiles` keeps its last `(inlinedFiles, hrefFiles)` on the Runner, and `restoreWorkspace()` replays them before the `updateFile` loop. The `io.files` snapshot is taken before the replay, since `provide_files` re-emits the file list from the worker, so a file the student edited afterwards still wins. Test: "hands provided files over again, since io.files omits the large ones".

src/frontend/state/Runner.ts
bmesuere: 🤖 Fixed in 587cdb25, with one deviation from the suggestion. `lintSource()` now awaits the backend in its own try, but a rejected launch makes it return `[]` without reporting or recovering rather than routing into the same catch: the launch failure is already reported once by whoever launched (`Papyros.launch()` or `recoverRuntime()`), and a launch error that matches a signature, such as the Safari 18.3 `Out of bounds memory access` inside `loadPyodide`, would otherwise retry the boot on every edit. This mirrors what `updateRunModes()` already does with launch failures. Test: "lints nothing after a failed launch instead of retrying it".

A failed relaunch still leaves the runtime dead until a reload, as before; the difference is that it no longer produces an uncaught error per lint on top.

TomNaessens (approved):

Issue #1141, opened 2026-09-04
Accessibility findings in the scratchpad editor, placeholder and input


TomNaessens:
Dodona scans its pages with pa11y (axe-core 4.11 + HTML_CodeSniffer, WCAG 2 AA) and keeps every page at zero findings. The coding scratchpad, which embeds `<p-code-runner>`, is the one page where findings had to be accepted instead of fixed, because they sit inside Papyros. This issue lists them so they can move upstream. Observed with @dodona/papyros 4.12.1 as shipped in Dodona; 4.12.2 does not touch the files involved. axe reports them at the shadow hosts, the targets below are the resolved shadow paths.

### 1. The code editor has no accessible name (`aria-input-field-name`, WCAG 4.1.2)

Target: `p-code-runner » p-code » p-code-editor » .cm-content`

CodeMirror renders the content element as `role="textbox" aria-multiline="true"`. `CodeEditor.ts` adds `aria-describedby="escape-hint"` through `EditorView.contentAttributes`, but nothing names the textbox, so a screen reader announces an unnamed multiline edit field. Adding a translated `aria-label` (for example "Code editor", or the file name in the multi-file layout) to that same `contentAttributes` facet fixes it.

**Fixed in #1142**: a new `accessibleName` property on the base editor sets `aria-label` through `contentAttributes`, translated ("Code editor" / "Code-editor"); the batch input and file editors get names too.

### 2. The editor placeholder fails contrast (`color-contrast`, WCAG 1.4.3)

Target: `p-code-runner » p-code » p-code-editor » .cm-placeholder`

The placeholder ("Write your Python code here and click 'Run' to execute...") keeps CodeMirror's base-theme colour `#888888`. Measured against the editor background in Dodona's two themes:

| theme | background | ratio | required |
| --- | --- | ---: | ---: |
| light | `#dbe1ec` | 2.69:1 | 4.5:1 |
| dark | `#50535c` | 2.16:1 | 4.5:1 |

[...]

PR #1142, opened 2026-09-04
Add a focus ring, accessible names and a visible escape hint to the editors

bmesuere:
This pull request fixes the accessibility gaps in the CodeMirror editors that an axe-core run and a manual audit turned up.

- The Material theme removed CodeMirror's focus outline without a replacement, so no editor showed that it had keyboard focus. Every editor now gets an inset 2px ring in the primary colour. (is tweaked in the later UI PR)
- No editor had an accessible name, so screen readers announced an unnamed text box. The code, batch input and file editors now set `aria-label` through a new `accessibleName` property on the base editor, wired into `EditorView.contentAttributes` the same way `placeholder` is, so it follows the locale.
- The "Press Escape followed by Tab" hint existed only for screen readers. It is now shown as a pill in the bottom-left corner while the editor is focused, the counterpart of the run-state pill on the right. (also restyle in the later PR)
- The placeholder colour uses the on-surface-variant token instead of CodeMirror's default `#888`, which failed the 4.5:1 contrast ratio in every theme.
- Test-code lines get a "T" gutter marker, like the debug and input line markers, so they are no longer distinguished by background colour alone.

Focused code editor with a focus ring and the escape hint pill

- Tab into the code editor: a ring appears around it and the hint shows bottom-left. Click elsewhere: both disappear.
- With VoiceOver on, focus the code editor, the batch input editor and a file editor: they are announced as "Code editor", "Input" and "File <name>".
- Switch the locale to Dutch and check the names follow.

Update #1141

### DIFF
--- a/src/frontend/components/code_mirror/CodeEditor.ts
+++ b/src/frontend/components/code_mirror/CodeEditor.ts
@@ -59,6 +59,7 @@ export class CodeEditor extends CodeMirrorEditor {
             :host {
                 width: 100%;
                 height: 100%;
+                position: relative;
             }
 
             .papyros-test-line {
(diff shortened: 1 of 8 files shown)


src/frontend/components/code_mirror/CodeEditor.ts
```diff
@@ -106,6 +107,25 @@ export class CodeEditor extends CodeMirrorEditor {
...
+                padding: 0.25rem 1rem;
+                overflow: visible;
+                clip-path: none;
+                white-space: nowrap;
+                font-size: 0.75rem;
+                border-radius: 1rem 1rem 0 0;
+                color: var(--md-sys-color-on-surface-variant);
+                background-color: var(--md-sys-color-surface-container);
+                z-index: 5;
+            }
```
bmesuere, line 128: this is restyled in a later PR

TomNaessens (approved):

src/frontend/components/code_mirror/CodeEditor.ts
bmesuere: 🤖 Confirmed, and nothing here needs changing: #1147 drops this `#escape-hint` block along with `position: relative` on the host, and moves the visible hint into the pane footer next to the run state. Left as is so this PR stays self-contained.

PR #1153, opened 2026-09-10
Stringify the input prompt before it reaches the frontend

bmesuere:
This pull request converts the prompt of `input()` and `prompt()` to a string before it is sent to the main thread, and stores it as a string in `InputOutput`.

Sentry DODONA-FRONTEND-2CE: `TypeError: this.content.cloneNode is not a function` in CodeMirror's `Placeholder.toDOM` while `BatchInput` renders.

**Cause.** `python_runner`'s `input()` forwards its argument unchanged and `papyros.py` sends it as the event data, so `input(5)` reaches the main thread as a number through pyodide's conversion and structured clone. `JavaScriptWorker.prompt(text)` does the same for `prompt(42)`. `BatchInput.placeholder` returns `io.prompt` when truthy and `CodeMirrorEditor` passes it to `placeholder()`, whose widget calls `cloneNode` on anything that is neither a string nor a function.

- `papyros.py`: `str(data["prompt"])`, matching what CPython's `input()` prints.
- `JavaScriptWorker.prompt`: `String(text)`.
- `InputOutput`: `this.prompt = String(e.data ?? "")` so the property matches its declared type whatever a backend sends.
- Tests: `prompt(42)` and `input(5)` leave `io.prompt` equal to `"42"` and `"5"`.

- Run `input(5)` in the Python sandbox: the input pane shows `5` as its placeholder instead of failing to render.
- Run `prompt(42)` in the JavaScript sandbox: same.

**Checklist**
- Tests: two cases in `InputOutput.test.ts`
- AI: Claude Code traced the Sentry event and wrote the change during dodona's weekly Sentry sweep

### DIFF
--- a/src/backend/workers/javascript/JavaScriptWorker.ts
+++ b/src/backend/workers/javascript/JavaScriptWorker.ts
@@ -50,7 +50,7 @@ export class JavaScriptWorker extends Backend {
     private prompt(text = ""): string {
         return this.onEvent({
             type: BackendEventType.Input,
-            data: text,
+            data: String(text),
             contentType: "text/plain",
         });
     }
--- a/src/backend/workers/python/papyros/papyros.py
+++ b/src/backend/workers/python/papyros/papyros.py
@@ -88,7 +88,7 @@ class Papyros(python_runner.PyodideRunner):
                         cb("output", data, contentType=part.get("contentType"))
             elif event_type == "input":
                 self._emit_turtle_snapshot()
-                return unwrap(cb("input", data["prompt"]))
+                return unwrap(cb("input", str(data["prompt"])))
             elif event_type == "sleep":
                 self._emit_turtle_snapshot()
                 return unwrap(cb("sleep", data["seconds"]*1000, contentType="application/number"))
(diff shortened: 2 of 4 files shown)


bmesuere:
🤖 Reviewed commit `b38d03de99753b92a31728adce343592e5aaa4b8`. I found no blocking issues in the numeric-prompt fix.

Validation: rebuilt the service worker and Python backend, then ran the input/output tests in Chromium. All 10 existing tests passed, including the new `prompt(42)` and `input(5)` cases. Type checking and ESLint for the changed TypeScript source files passed. All current CI checks pass. An additional browser test passed for JavaScript prompts using `0`, `false`, `null`, `undefined`, and an object with a custom `toString()`.

**Non-blocking follow-up:** Python `input(False)` still fails before it reaches the new conversion. `python_runner.input` first calls `self.output("input_prompt", prompt)`, and `BackendEventQueue.put` passes the boolean to `TextDecoder.decode`, which raises a `TypeError`. I reproduced the same failure with the previous Python callback restored, so this is an existing limitation rather than a regression from this PR. Converting the Python prompt before that output call would cover this case too. Python `input(0)` passed the additional check.

TomNaessens (approved):

PR #1158, opened 2026-09-13
Replace pylint with ruff for linting Python

bmesuere:
**TODO (author) — drafted by Claude: proofread, verify and remove this line.**

This pull request replaces pylint with ruff as the linter for Python. It is the full switch that #1157 spikes, without the spike's toggle, timings and report, so the diff against `main` shows what a real migration adds and removes. It branches from `main`, not from #1157, and the two are not meant to be merged together: this one supersedes the spike if we go ahead.

Ruff runs from `@astral-sh/ruff-wasm-web` 0.16.7 inside the existing Python worker, loading in parallel with Pyodide. A lint no longer installs the code's imports first, and the Python tarball no longer ships pylint, astroid and their dependencies. Diagnostics carry ruff's rule id, which the lint panel shows under the message (same rendering as in #1157):

The demo editor: `import re` and `x = undefined_name` are underlined, and the lint panel lists "`re` imported but unused / F401" and "Undefined name `undefined_name` / F821"

## Size of the change

| | added | removed |
|---|---:|---:|
| Python side (`linting.py`, `pylint_ast_checker.py`, `pylint_turtle_brain.py`, `pylint_config.rc`, two lines of `papyros.py`, two lines of `requirements.txt`) | 0 | 527 |
| TypeScript (`ruff.ts` new, `PythonWorker.ts`, `CodeEditor.ts`, `Code.ts`, `Backend.ts`, `vite.config.ts`) | 203 | 124 |
| tests (`RuffDiagnostics.test.ts` new, `Runner.test.ts`) | 55 + 21 | 52 |
| total, excluding `yarn.lock` | 204 | 651 |

Python tarball on disk: 454 KB, was 1,532 KB. Unpacked in the worker's file system: 2.4 MB, was 8.1 MB. The ruff wasm that comes in instead is 10.9 MB raw, 2.6 MB brotli, fetched once and cached; see the payload table in #1157 for the comparison against pylint's share of the tarball and against Pyodide itself.

## Removed

- `src/backend/workers/python/papyros/linting.py`: the pylint runner, the `--msg-template` re-parser, the `SEVERITIES` table, the E0001 unwrapping that hid the temp file name from syntax errors, and the astroid monkeypatch that stubbed stdlib modules astroid cannot build under Emscripten (`os` pulls in `posix`).
- `src/backend/workers/python/papyros/pylint_ast_checker.py`: papyros' own pylint checkers, copied from judge-pythia. `NoOpChecker` (C0001 to C0006: useless `pass` in an `else`, `if x == True`, `x += 0`, `x += ''`, `x *= 1`, `x = x`), `RewriteChecker` (C0007 `if: pass else:`, C0008 `x = x + 1`) and `UnnecessaryConversionChecker` (C0009). Of these only `x == True` (ruff E712) and `x = x` (ruff PLW0127) have a ruff equivalent; the rest are gone. If we want them back they have to be written as ruff has no plugin system: either as a small AST pass in Pyodide or upstream.
- `src/backend/workers/python/papyros/pylint_turtle_brain.py`: an astroid brain so pylint did not report false positives on the bundled turtle module. Not needed without astroid.
- `src/backend/workers/python/papyros/pylint_config.rc`: the rule configuration. Its `disable` list is mirrored in `RUFF_OPTIONS`, see below.
- `Papyros.lint()` in `papyros.py`, which was the only caller of `linting.py`.
- `pylint==4.0.7` and `tomli==2.4.1` from `requirements.txt`. tomli was only there for pylint; nothing else in the tarball imports it (`grep -rn "import tomli"` over the unpacked tarball finds only pylint). With them go astroid, isort, dill, mccabe, tomlkit, platformdirs and the stray macOS `.so` from the tomli wheel that #1157 noted.
- In `PythonWorker.lintCode()`: the `installImports(code)` call before every lint. `installImports()` itself stays, `runCode()` still needs it.
- In `CodeEditor.ts`: `reLintOnLoaded`, the `forceLintEffect` state effect, the `needsRefresh` option on the linter, the `papyros` setter with its subscription to `BackendEventType.Loading`, and the imports that only served them (`forceLinting`, `BackendEvent`, `BackendEventType`, `parseData`, `Papyros`). This machinery re-ran the lint when a package finished installing, to clear pylint's stale "Unable to import X" while the wheel was still downloading. Ruff never resolves imports, so there is nothing to re-lint for. The `.papyros=${this.papyros}` binding in `Code.ts` goes with the setter; the editor does not need the instance for anything else.

## Added

- `src/backend/workers/python/ruff.ts`: `RUFF_OPTIONS` and `toWorkerDiagnostics()`, identical to #1157. Positions: ruff reports 1-based rows and columns with an exclusive end, in UTF-16 units because the workspace is created with `PositionEncoding.Utf16`; the mapping subtracts one from both columns and leaves the rows. Severity by rule prefix: syntax errors, `F`, `E9`, `PLE` are errors; `W`, `PLW`, `B` warnings; the rest info. Ruff has no severity of its own, so this table is ours to tune. A syntax error is `code: "invalid-syntax"` in 0.16.7 (the typings say `null`); both are handled and produce an error without a `code`.
- `PythonWorker`: a `ruff: Promise<Workspace>` started in `launch()` before the Pyodide boot and awaited by `lintCode()`. `loadRuff()` calls ruff's default `init()`, which fetches `ruff_wasm_bg.wasm` relative to `ruff_wasm.js` via `import.meta.url`, then constructs the workspace.
- `WorkerDiagnostic.code?: string` in `Backend.ts`; `CodeEditor.ts` passes it as `source` to CodeMirror, which renders it under the message in the lint panel.
- `vite.config.ts`: `optimizeDeps.exclude: ["@astral-sh/ruff-wasm-web"]`, so the dev server serves the package in place and the wasm is found next to its script.
- `package.json`: `@astral-sh/ruff-wasm-web` pinned to `0.16.7`; the wasm API is marked experimental.

## Rule set

```ts
{
    "target-version": "py314",
    "line-length": 120,
    lint: {
        select: ["E", "W", "F", "PL"],
        ignore: ["E501", "W291", "W292", "E402", "PLR0904", "PLR0911", "PLR2004"],
        "dummy-variable-rgx": "^_$",
    },
}
```

This mirrors the rc's `disable=W0311,W0621,W0622,R0902,R0903,C0114,C0115,C0116,C0301,C0303,C0304,C0413,I0011` where ruff has the rule: C0301 is E501, C0303 is W291, C0304 is W292, C0413 is E402, R0904 is PLR0904, R0911 is PLR0911. R0902 and R0903 are preview-only in ruff (PLR0902, PLR0903) and cannot be named without preview mode, so they are absent rather than ignored. W0311 (bad indentation), W0621 (redefined outer name), W0622 (redefined builtin) and C0114 to C0116 (docstrings) have no rule in the selected families, so nothing needs ignoring. PLR2004 (magic value comparison) is ignored on top because it fires on every `if n > 10:` a beginner writes. The rc's relaxed naming regexes have no counterpart: naming rules (`N`) are not selected.

What this rule set does not report that pylint did, from the diff in #1157: wrong argument count to your own function, a method or attribute that does not exist (`xs.push(3)`, `s.uppercase()`), a wrong name from a module (`from math import sqroot`), an unresolvable import, a method without `self`, unreachable code, `if True:`, `range(len(xs))`, disallowed names, and the custom checkers listed above. It reports E703 (semicolon) and W293 (whitespace on a blank line) that pylint's configuration did not, and both F704 and PLE1142 for `await` outside an async function. Tuning the families and ignores is a follow-up; this PR keeps the spike's set so the two PRs agree.

## Tests

- `test/__tests__/RuffDiagnostics.test.ts` (new): the column shift and `code` on a normal diagnostic, the syntax error mapping for both `"invalid-syntax"` and `null`, and the severity table.
- `test/__tests__/state/Runner.test.ts`: five pylint tests are removed because what they tested no longer exists: "should lint bare import re", "should lint code that uses pandas without hanging or a false import-error", "should report an import-error for a genuinely missing module", "should not flag stdlib modules astroid cannot build (os) as unimportable", "should report a wrong name imported from a stubbed stdlib module (os)". One test replaces them: linting `import sympy` resolves in under a second, returns no diagnostics and leaves `loadingPackages` empty, where pylint would have downloaded sympy first. The three assertion tests are rewritten for ruff: an undefined name comes back as an error with code `F821` on line 2, columns 4 to 18 (0-based); `print 'hello'` comes back as errors without a code; `x = 1; y = 2` comes back as one info diagnostic with code `E702` (pylint's version of this test used a function without a docstring and expected the missing-docstring info, which the rc disabled, so it actually asserted on `disallowed-name` for `foo`).
- The test count goes from 202 on `main` to 195.

## Verification

- `yarn setup`, `yarn typecheck`, `yarn lint`, `yarn test`: green, 26 files, 195 tests.
- `yarn build:lib`: succeeds; `dist/backend/workers/python/` has the rebuilt tarball, `ruff.js` and `PythonWorker.js` with the bare `@astral-sh/ruff-wasm-web` import for the consumer's bundler; `dist/InputServiceWorker.js` present.
- The Dodona build question is answered in #1157: its Vite 8.2.2 build emits the wasm as a hashed asset for the `new URL(..., import.meta.url)` idiom, the same way it already handles the tarball, so no extra parameter is needed. Not verified against a real Dodona build.

- `yarn setup`, `yarn start`, open the demo, pick Python.
- Type `import re` and on the next line `x = undefined_name`. Both lines get a red gutter dot and underline; Cmd-Shift-M (Ctrl-Shift-M on Linux) opens the lint panel with `F401` and `F821` under the messages.
- Type `import pandas as pd` alone: no "loading pandas" state appears and the line is underlined as an unused import. Add `print(pd)` and run: the package downloads at run time as before.
- Type `print 'hello'`: one or more errors, no rule id.
- Pick the Erroneous example: the syntax error on line 6 is underlined; lines 17 and 27 get an info mark for the semicolon.

**Checklist**
- Tests: see above.
- Accessibility: the lint panel gains the rule id line under each message; no other markup or colour changes.
- Docs: the README does not mention pylint; n/a.
- Announcement: ✨ The Python scratchpad lints with ruff instead of pylint: diagnostics appear without waiting for the code's packages to download, and each message shows its ruff rule id.
- AI: Claude Code implemented the switch and the tests based on the spike in #1157; Bart reviewed.


bmesuere:
- [ ] figure out the wasm url
- [ ] code doesn't need to be shown
- [ ] python version target in wasm should be the same as pyodide. github action check?
- [ ] markdown parsing of messages?

src/backend/workers/python/PythonWorker.ts
```diff
@@ -41,6 +49,7 @@ export class PythonWorker extends Backend {
         allowJspi: boolean = true,
     ): Promise<void> {
         await super.launch(onEvent, pyodideAssetURL, allowJspi);
+        this.ruff = PythonWorker.loadRuff();
```
bmesuere, line 52: `PythonWorker` is maybe not the best name anymore

